PR #4, opened 2024-10-31
Add account requirements

tchojnacki:


mlodybercik (approved):
>> Aren't three questions in user stories necessary? _Whom?_, _What?_ and _Why?_

> It does not matter.

All of my complaints have been resolved via a private message.

LGTM 🙆‍♀️

PR #8, opened 2024-11-05
User stories and UI mockups

mlodybercik:
- user stories done, UI mockup WIP
- UI mockups done
- Added images to document

--- 

Changes to discuss:

- add lat/long to `Stop` model,
- add resolved flag to `Accident`,
- add exception to screen compatibility for system admin interfaces.


documentation/course/E1.md
```diff
@@ -134,7 +134,13 @@ W zakres projektu wchodzi:
...
 ### Logistyka (Przemysław Barcicki)
 
-- `LOG/01` - TODO@mlodybercik - historyjki użytkownika
+- `LOG/01` - Jako _administrator_ chcę mieć możliwość przeglądania, dodawania, edycji i usuwania _linii_, aby zarządzać trasami _pojazdów_.
+- `LOG/02` - Jako _administrator_ chcę mieć możliwość przeglądania, dodawania, edycji i usuwania _przystanków_, aby zarządzać miejscami, w których _pasażerowie_ mogą wsiąść lub wysiąść z _pojazdów_.
+- `LOG/03` - Jako _administrator_ chcę mieć możliwość przeglądania, dodawania, edycji i usuwania _rozkładów jazdy_, aby zarządzać godzinami odjazdów _pojazdów_ z _przystanków_.
+- `LOG/04` - Jako _administrator_ chcę mieć możliwość przeglądania, dodawania, edycji i usuwania _pojazdów_, aby zarządzać dostępnymi środkami transportu.
+- `LOG/05` - Jako _kierowca_ chcę mieć możliwość przeglądania, zgłaszania i edycji _awarii_, aby informować _pasażerów_ o problemach na _kursie_.
+- `LOG/06` - Jako _bileter_ chcę mieć możliwość przeglądania, zgłaszania i edycji _awarii_, aby informować _pasażerów_ o problemach na _kursie_.
+- `LOG/07` - Jako _administrator_ chce mieć możliwość przeglądania _awarii_, aby monitorować sytuację na _trasach_.
```
tchojnacki: **nit:** As mentioned in chat, I'd split them up (e.g. CRUD becomes five requirements), but it's up to you honestly, I won't block the PR because of it.

documentation/course/E1.md
```diff
@@ -150,7 +156,7 @@ Wymagania niefunkcjonalne zostały podzielone zgodnie z metodyką FURPS+ (Functi
 ### Użyteczność
 
 1. System powinien być zaprojektowany tak, aby 90% użytkowników mogło wykonać podstawowe czynności, takie jak zakup biletu czy sprawdzenie rozkładu jazdy, bez konieczności korzystania z instrukcji lub pomocy technicznej.
-2. System powinien być zoptymalizowany zarówno dla urządzeń desktopowych, jak i mobilnych, oferując pełną funkcjonalność na ekranach o rozdzielczościach od 320x480 pikseli.
+2. System powinien być zoptymalizowany zarówno dla urządzeń desktopowych, jak i mobilnych, oferując pełną funkcjonalność na ekranach o rozdzielczościach od 320x480 pikseli (wyjątkiem od reguły są interfejsy administratora).
```
tchojnacki, line 201: Sounds good 👍

documentation/course/E1.md
```diff
@@ -210,6 +216,7 @@ classDiagram
     class Accident {
         long id [1] #lcub;id, readonly#rcub;
         Instant time [1] #lcub;readonly#rcub;
+        Boolean resolved [1] = false
```
tchojnacki: **nit:** `Boolean` suggests that `resolved` could be `null` and I don't see any business reason for that to be the case.
```suggestion
        boolean resolved [1] = false
```

documentation/course/E1.md
```diff
@@ -269,6 +276,8 @@ classDiagram
     class Stop {
         long id [1] #lcub;id, readonly#rcub;
         String name [1] #lcub;readonly#rcub;
+        BigDecimal latitude [1] #lcub;readonly#rcub;
+        BigDecimal longitude [1] #lcub;readonly#rcub;
```
tchojnacki: **nit:** Very minor, but I think that there's little benefit to using `BigDecimal` here instead of `double`. It was obviously required for money, but if we misplace the stop a few centimeters in any direction it isn't a problem.

documentation/course/E1.md
```diff
@@ -612,16 +627,89 @@ TODO@piterek130
...
+- `LOG/02/04` - Usuwanie przystanku.
+
+![LOG/02/04](images/Stop4.png)
+
 ### Wymaganie 8
 
-TODO@mlodybercik
+- `LOG/06` - Jako _bileter_ chcę mieć możliwość przeglądania, zgłaszania i edycji _awarii_, aby informować _pasażerów_ o problemach na _kursie_.
 
 #### Mapa nawigacyjna
```
tchojnacki: Let's keep the formatting consistent with @jakubzehner. I don't really care which format we will end up using, but currently we are using two clashing standards.

I'd even say that your formatting looks better than Kuba's. In that case you might change how requirements are written down in his part.

I'd also write down your name next to the requirement, so that it will be easier for our instructor to check these, since they are treated as individual work.

documentation/course/E1.md
```diff
@@ -612,16 +627,89 @@ TODO@piterek130
...
 
+```mermaid
+classDiagram
+    direction LR
```
tchojnacki, line 1031: Good job 👍

tchojnacki (approved):

documentation/course/E1.md
jakubzehner: ahh yes, Java

documentation/course/E1.md
jakubzehner: I agree

documentation/course/E1.md
mlodybercik: https://stackoverflow.com/a/1196429

Fair, it is not critical in any sense. It reduces the size of a row by few bytes. We don't need the resolution of floats, but also we don't plan to do any operations on geospatial data. Its purely for administrators convenience. 

We'll go with floats then.

documentation/course/E1.md
mlodybercik: I assumed that for simplicity, every one of us would take tightly-linked requirements for UI prototyping, thus not requiring multiple navigational maps. @jakubzehner has 2 + 1 + 1 by the looks of it, so my formatting as is, will not cut it. I'll try to come up with something that will work for us both.

jakubzehner (approved):
LGTM

piterek130 (approved):
LGTM

PR #12, opened 2024-11-07
Add navigation maps

jakubzehner:


documentation/course/E1.md
```diff
@@ -708,37 +708,180 @@ classDiagram
...
+    class EkranZakupuBiletu {
+        <<ServerPage>>
+        }
+
+    PasekNawigacyjny ..> EkranTwoichBiletow: «Reloads»\n{ event = btnBilety }
+    PasekNawigacyjny ..> EkranOfertyBiletow: «Reloads»\n«StartUseCase»\n{ event = btnKupBilet }
+    EkranOfertyBiletow ..> PasekNawigacyjny: «Navigates»\n«EndUseCase»\n{ event = dowolnyPrzycisk }
+
+    EkranOfertyBiletow ..> EkranZakupuBiletu: «Navigates»\n«EndUseCase»\n{ event = btnKup }
+    EkranZakupuBiletu ..> EkranOfertyBiletow: «Navigates»\n«StartUseCase»\n{ event = btnPowrot }
```
piterek130, line 800: Is this good that the same button navigates to diffrent pages? (btnPowrot)

piterek130 (approved):
One uncertainty thing but generally LGTM.

documentation/course/E1.md
jakubzehner: I don't know how it would be any different, `Powrot` goes back to the previous screen, so it depends from which one you entered. Having another button would just make the frontend implementation more difficult.

PR #31, opened 2024-12-17
PR: Scenario sequence diagrams for ticket purchase and inspection

jakubzehner:


jakubzehner:
I wonder if we should add alt block for login sequence where login data is invalid and we do not create authorization token @tchojnacki

documentation/course/e2/images/sequence-diagram-scenario-ticket-inspection.drawio.svg
mlodybercik, line 1: >display ticket information

It sounds funny, and arrow in the direction of JakPrzyjade implies that Passenger is displaying the code to our system. How about "get current ticket"? It's not bad, but it could be better I guess.

> handing out the fine

Shouldn't "handing" be infinitive? Every other verb is. How about: "Hand out the fine"?

documentation/course/e2/images/sequence-diagram-scenario-ticket-purchase.drawio.svg
mlodybercik, line 1: Shouldn't self-call be partially overlapping the JakPrzyjade lifeline?

<sub>**nittest nit of all nits**: Payment Gateway is slimmer than JakPrzyjade</sub>

mlodybercik (changes requested):
1hf37d

mlodybercik (approved):
istockphoto-157030584-612x612

tchojnacki:
> I wonder if we should add alt block for login sequence where login data is invalid and we do not create authorization token @tchojnacki

Yes, as discussed, I'll add it.

documentation/course/e2/images/sequence-diagram-scenario-ticket-inspection.drawio.svg
tchojnacki, line 1: It is probably invalid UML-wise, but it gets the point across, I like it.

Should the lines returning to the Passenger have some labels?

documentation/course/e2/images/sequence-diagram-scenario-ticket-purchase.drawio.svg
tchojnacki, line 1: The return line from `charge wallet` seems unnecessary if we don't return anything. Note that activations without returns are valid UML:

sequence-execution-spec-self
seq04

tchojnacki, line 1: This diverges from the prototype we had, where user had a payment option dropdown and had to explicitly select credit card to be charged. Is that on purpose?

jakubzehner: done

jakubzehner: resolved as discussed

documentation/course/e2/images/sequence-diagram-scenario-ticket-inspection.drawio.svg
jakubzehner: resolved as discussed

tchojnacki (approved):

PR #32, opened 2024-12-18
PR: E2 package diagram for Jobberknoll and Feather

tchojnacki:


documentation/course/e2/images/sequence-diagram-mechanism-12.drawio.svg
tchojnacki, line 1: Changed `int` to `ext`, as noticed by Przemek. I also removed the `{SRV} Service` to emphasise, that the diagram ends at the moment of that message call.

The notation with arrow aiming into black dot has the following meaning:

> A lost message can occur when a message is sent to an element outside the scope of the UML diagram. A lost message contains a dot at the end of the arrowhead to indicate the destination is unknown.

tchojnacki:
TODO @tchojnacki: Update deployment diagram.

documentation/course/e2/images/sequence-diagram-mechanism-12.drawio.svg
mlodybercik: I've also noticed that communication `hogwart` -> `phoenix` can be different than plain `GET`. Maybe this possibility should also be reflected in the diagram?
`{METHOD} phoenix/api/{SRV}/{RES}`
...
`{METHOD} {srv}/ext/{RES}`

documentation/course/e2/README.md
```diff
@@ -2011,7 +2011,86 @@ TODO @everyone
...
+4. **znajomość języka** w zespole realizującym podsystem oraz poza systemem (przez co najmniej jedną osobę spoza zespołu) - wymóg wymieniony powyżej, który ma na celu zapewnienie sprawnego i bezstronnego code review;
+5. **dostępność algebraicznych typów danych** - w samym centrum modelu informacyjnego podsystemu znajduje się typ stanowiący algebraiczny typ danych, a dokładniej rekord z wariantami (ang. _tagged union_ lub _sum type_), w związku z tym istotne jest, aby język wspierał takie konstrukcje natywnie, bez konieczności emulacji za pomocą mechanizmów programowania obiektowego.
+
+Z uwagi na powyższe wymagania, odpowiedni okazał się być język TypeScript, spełniający wszystkie wymagania. W przypadku wykorzystania języka TypeScript lub JavaScript na serwerze, konieczny jest wybór środowiska uruchomieniowego, gdzie dwie najpopularniejsze opcje to Node.js oraz Deno. Silnik Deno został stworzony przez autora Node.js jako próba naprawienia problemów, które pojawiły się przez lata w starszym Node.js. Względem Node.js, Deno:
```
mlodybercik: In the span of 3 sentences, you used Node.js 4 times. Thats a new high score! 🎉

documentation/course/e2/images/deployment-diagram.drawio.svg
tchojnacki, line 1: **note:** I know that we might want to roll with something like Alpine instead of Ubuntu for the actual prod deployment, but I wanted to avoid potential problems with common packages being unavailable (e.g. OpenSSL, CURL).

mlodybercik: Alpine supports them all.

https://pkgs.alpinelinux.org/package/edge/community/aarch64/deno
https://pkgs.alpinelinux.org/package/edge/main/aarch64/openssl
https://pkgs.alpinelinux.org/package/edge/main/aarch64/curl

mlodybercik: Better yet, Deno has its own Docker recipes: https://github.com/denoland/deno_docker

jakubzehner (approved):
Overall it looks good, I have no specific comments and all regarding the package diagram we have already discussed in a private message.

My only concern is whether we should consider non-architectural patterns in the architecture planning stage. Such as the behavioral design patterns you mentioned, e.g. “Command” or “Chain of responsibility”, or even structural patterns such as “Adapter”

documentation/course/e2/images/package-diagram-jobberknoll.drawio.svg
mlodybercik, line 1: nit: Some scaling ratios cause lines under other lines to be visible.
obraz

tchojnacki: Fixed 👍

mlodybercik (approved):
obraz

documentation/course/e2/images/deployment-diagram.drawio.svg
tchojnacki: Alright, changed to Alpine.

I knew about Deno images and didn't pick them on purpose. Deno supports compilation to standalone executables, so after producing the artifact I don't need the runtime on the production machine. Note the 4th point under Node vs Deno, and also how I've changed `jobberknoll.js` to `jobberknoll` and `feather.js` to `feather`.

Ideally we'd want a multi-stage Dockerfile, where the first stage uses Deno containers to compile the app and the second stage is as bare-bones as possible and just runs the app. I had something similar in the Cloud course project and it worked flawlessly (some commands omitted for brevity):
```Dockerfile
FROM ubuntu:24.10 AS run
WORKDIR /app
COPY --from=build ./app/server ./server
CMD ["./server"]
```

tchojnacki:
@jakubzehner 
> My only concern is whether we should consider non-architectural patterns in the architecture planning stage. Such as the behavioral design patterns you mentioned, e.g. “Command” or “Chain of responsibility”, or even structural patterns such as “Adapter”

I think so, these patterns are not added on top of the project structure, they are the reason for such structure. If not for the Strategy pattern, you wouldn't split `IUserRepository` and `PostgresUserRepository` across modules. If not for the Command pattern, you wouldn't have a `use-cases` folder and would instead have a `user-service.ts` file, etc.

PR #42, opened 2024-12-22
PR: Implement the minimum viable Jobberknoll prototype

tchojnacki:
obraz
obraz
obraz
obraz
obraz


mlodybercik (approved):
As discussed in PMs.
christmas-cat

PR #43, opened 2024-12-23
PR: Implement "get account" endpoint for internal consumers

tchojnacki:
Will get these integrated into CI/CD soon:
obraz
obraz
obraz

OpenAPI:
obraz


implementation/jobberknoll/api/shared/mappers/account-mapper.ts
```diff
@@ -0,0 +1,12 @@
+import type { Account } from "@jobberknoll/core/domain";
+import type { AccountDto } from "~/shared/contracts/account-dto.ts";
+
+export function mapAccountToDto(account: Account): AccountDto {
+  return ({
+    id: account.id,
+    type: account.type,
+    fullName: account.fullName,
+    email: account.email,
+    phoneNumber: "phoneNumber" in account ? account.phoneNumber : undefined,
```
jakubzehner, line 10: Can't this be replaced by a simple `phoneNumber: account.phoneNumber ?? undefined`?

documentation/contracts.md
```diff
@@ -44,14 +44,80 @@ While a more detailed response could look like this:
...
+The JSON objects returned alongside `4xx` and `5xx` HTTP status codes should follow the following schema:
+
+```TypeScript
+type AppError = {
+  code: number; // const literal
+  kind: string; // const literal, lowercase kebab-case, without trailing "-error"
+  messageEn: string;
+  messagePl?: string; // optional, required for user-facing errors
+};
+```
```
jakubzehner: If I understand correctly, this contract applies to all services not just yours. That's why I think we could choose some other language than TypeScript to represent dto, because this one is very imprecise and, for example, does not distinguish between int and float.

jakubzehner: (I'm not saying it has to be Java)

documentation/contracts.md
```diff
@@ -44,14 +44,80 @@ While a more detailed response could look like this:
...
+  </tr>
+  <tr>
+    <td><code>404</code></td>
+    <td><code>account-not-found</code></td>
+  </tr>
+</table>
+
+## Pagination
+
+Pagination follows the [Spring Boot conventions](https://docs.spring.io/spring-data/rest/reference/paging-and-sorting.html). The query parameters are:
```
jakubzehner: ❤️

implementation/jobberknoll/api/int/int-controller.ts
```diff
@@ -1,17 +1,32 @@
...
 import type { Controller } from "~/shared/controller.ts";
 import { configureDocs } from "~/shared/docs.ts";
-import { getHealthHandler, getHealthRoute } from "./routes/get-health-route.ts";
+import { configureErrorHandler, defaultHook } from "~/shared/hooks.ts";
+import {
+  getAccountByIdHandler,
+  getAccountByIdRoute,
+  getHealthHandler,
+  getHealthRoute,
+} from "./routes/mod.ts";
```
jakubzehner, line 12: Are you going to import every route and handler from this file to “register” it here?

I'm not saying that importing it from individual files would be better, because it wouldn't. But here imports will quickly take up half the volume of the file. I don't have an idea of a better solution right now, but some probably exists.

implementation/jobberknoll/api/shared/contracts/account-dto.ts
```diff
@@ -0,0 +1,36 @@
...
+}).openapi("AccountDto", {
+  description: "Account details.",
+  examples: [{
+    id: uuid(),
+    type: "passenger",
+    fullName: "John Smith",
+    email: "john.smith@example.com",
+    phoneNumber: "+48 123 456 789",
+  }],
+});
```
jakubzehner, line 34: Tbh it looks like developer's hell to me and I wouldn't push myself into it, but it's fine, it's your hell, so I don't bother.

However it looks like you could handle most of the data validation here, unless you really want to do it in `UseCase` which is what the `TODO` you left behind suggests.

jakubzehner: Ofc I understand that this is an “output” dto so in this case there is nothing to validate

implementation/jobberknoll/api/shared/openapi.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  const testError = {
+    code: 400,
+    kind: "test2",
+    messageEn: "Test error message.",
+  };
+
+  const result = TestDto.safeParse(testError);
+
+  assert(!result.success);
+});
```
jakubzehner: How about doing some parameterized testing here?

jakubzehner: If this is even possible in your framework xd

implementation/jobberknoll/core/domain/entities/account.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export type Admin = AccountBase & {
+  type: "admin";
+};
+
+export type Inspector = AccountBase & {
+  type: "inspector";
+};
+
+export type Account = Passenger | Driver | Admin | Inspector;
```
jakubzehner, line 29: This somewhat relates to your comment at my place. In my opinion, we should add a `Guest` here. But we should discuss this in a larger group.

implementation/jobberknoll/core/shared/option.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export type Option<T> = Some<T> | None;
+
+export const some = <T>(value: T): Some<T> => ({ tag: "some", value });
+export const none = (): None => ({ tag: "none" });
+
+export const isSome = <T>(option: Option<T>): option is Some<T> =>
+  option.tag === "some";
+export const isNone = <T>(option: Option<T>): option is None =>
+  option.tag === "none";
```
jakubzehner, line 12: Perhaps you should do it in rust after all? xD

implementation/jobberknoll/core/shared/uuid.ts
```diff
@@ -0,0 +1,19 @@
+import { generate, validate } from "@std/uuid/unstable-v7";
```
jakubzehner: sus

implementation/jobberknoll/setup.ts
```diff
@@ -1,11 +1,24 @@
 import { buildApp, IntController } from "@jobberknoll/api";
+import { buildService } from "@jobberknoll/app";
+import type { Account } from "@jobberknoll/core/domain";
+import { MemoryAccountRepo } from "@jobberknoll/infra";
 
 export function setupProd() {
-  const intController = new IntController();
+  // TODO @tchojnacki: Replace with a persistent repository
```
jakubzehner, line 7: 👍

implementation/jobberknoll/api/shared/mappers/account-mapper.ts
tchojnacki: Nope `.` works for possibly-`undefined` values, but not possibly-nonexistent values, and `phoneNumber` is not `undefined` in other `Account` types, it is missing at all.

obraz

(to avoid the next question, `?.` wouldn't work either)

documentation/contracts.md
tchojnacki: > If I understand correctly, this contract applies to all services not just yours.

Correct.

> that's why I think we could choose some other language than TypeScript to represent dto

Will fix :+1:

implementation/jobberknoll/api/int/int-controller.ts
tchojnacki: I'll see how far it takes me, if it becomes too long I might just `import * as r from "./routes/mod.ts";`, and then `r.getHealthRoute`.

implementation/jobberknoll/api/shared/contracts/account-dto.ts
tchojnacki: > Tbh it looks like developer's hell to me

Is it really much worse than [writing a 1.3k line JSON file by hand](https://github.com/RustyRipper/online-city-tickets/blob/main/online-city-tickets-backend/src/main/resources/static/openapi.json)?

> However it looks like you could handle most of the data validation here

That's what happens, supported by `defaultHook` in `/api/shared/hooks.ts`. Both bodies, path parameters, query parameters and headers will (or actually are) automatically checked in accordance to the schema.

In fact, apart from validating request data at runtime, it validates my code at compile-time. I can't return data non-conforming to the OpenAPI schema without a compilation error. If I were to change `getAccountByIdHandler` in `/api/int/routes/get-account-by-id-route.ts` to return `201` instead of `200`, the following thing happens:
obraz

implementation/jobberknoll/api/shared/openapi.test.ts
tchojnacki: I'm not even using a testing framework, `Deno.test` is a built-in testing primitive. I could import a whole framework but honestly I don't feel the need. I can make it parametrized by just wrapping in a good old `for` loop (like in `/tests/docs-e2e.test.ts`).

Will do 👍

implementation/jobberknoll/core/domain/entities/account.ts
tchojnacki: It's a domain type meant to be used for business logic and later the persistence/data-access layer is built upon it. So I don't see the need to hold `Guest` here, since guests don't exist in the database (and neither in the app state, since our services should be stateless).

I'm guessing you thought it was a role enumeration? As in:
```
type Role = "guest" | "passenger" | "driver" | "admin" | "inspector";
```

I don't have such type yet, but if I will, it would contain a `guest` literal 👍

implementation/jobberknoll/core/shared/option.ts
tchojnacki: I should have 😭 

I kind of wanted to do error handling through exceptions, but:

1. They are slower, because of all the stack unwinding, etc. (I know it would be negligible for our use case).
2. I don't like using exceptions for control flow, they should ideally be, as the name suggests, for exceptional occurrences, and while I would classify an internal database error as such, request with an invalid ID doesn't really fill that definition for me.
3. Due to how JS works, `catch` blocks in TypeScript can only use the `any` and `unknown` types https://oida.dev/typescript-typing-catch-clauses, so there would be a lot of mindless (and unsafe) typecasting.

Right now, all sensible errors are handled with algebraic data types, and others (e.g. developer error, database failure) cause a `HTTP 500` on the request.

implementation/jobberknoll/core/shared/uuid.ts
tchojnacki: Since it's a part of the standard library, there are strict stabilization guidelines.

As an alternative, I can import something like `npm:uuid`, but I'm guessing that an unstable lib from runtime authors is better than a stable but entirely community-supported lib.

implementation/jobberknoll/api/shared/contracts/account-dto.ts
tchojnacki: @jakubzehner totally forgot about the second part of the question:

> unless you really want to do it in UseCase which is what the TODO you left behind suggests.

This place only support dependency-less, synchronous validation. So I can check whether an email has valid format here, but not whether it is unique (since it requires a call to the DB). So I might end up doing format validation here and the rest in use cases.

implementation/jobberknoll/core/shared/option.ts
jakubzehner: I understand your pain, throwing exceptions is the worst thing I do in my service implementation, but in my case it's standardized enough that I don't undertake to do everything through types.

tchojnacki:
TODO @tchojnacki:
- parametrize `openapi.test.ts` tests
- change `AppError` definition in `contracts.md`
- document established HTTP Headers in `contracts.md`
- switch to a stable UUIDv7 provider

implementation/jobberknoll/core/shared/uuid.ts
jakubzehner: "unstable"
image

implementation/jobberknoll/core/shared/option.ts
tchojnacki: @jakubzehner .NET 8.0 solved it nicely:
- https://nikiforovall.github.io/dotnet/2024/08/17/result-endpoints.html
- https://fast-endpoints.com/docs/get-started#union-type-returning-handler

We will see it in Spring too, probably in 5-10 years.

implementation/jobberknoll/core/shared/uuid.ts
tchojnacki: Sure, will change 👌

jakubzehner: If you say that you trust this "unstable" more than from the community, then fine, it can stay

documentation/contracts.md
```diff
@@ -42,16 +44,188 @@ While a more detailed response could look like this:
...
+
+Invalid pagination request (negative `page`, size `smaller` than 1, `size` bigger than 100) SHOULD return a `422` error with the `schema-mismatch` kind.
+
+## Endpoints route
 
 TODO
 
 ## Headers
 
+The well-known headers of the system are:
```
tchojnacki, line 188: cc: @mlodybercik @piterek130

documentation/contracts.md
```diff
@@ -42,16 +44,188 @@ While a more detailed response could look like this:
...
+```Java
+@Data
+@AllArgsConstructor
+public class AppError {
+    @NotNull private final int code;
+    @NotNull private final String kind;
+    @NotNull private final String messageEn;
+    private final String messagePl;
+}
+```
```
jakubzehner: Nit: in the documentation, I would not use the lombok annotation, and maybe even use record here instead of class

jakubzehner (approved):
One minor comment, other than that everything seems fine. Good job 👌

documentation/contracts.md
jakubzehner: Also, `int` can't be null so `@NotNull` is not necessary here, if not even invalid

PR #45, opened 2024-12-26
PR: Clabbert - enhanced architecture with mediator, error handling, user header interceptor, and CI integration

jakubzehner:
This pull request introduces key improvements to the project's architecture and development process:
- Implementation of the [Mediator](https://refactoring.guru/design-patterns/mediator) design pattern to streamline communication between components
- Addition of custom error handling using for centralized error management - `ApiExceptionHandler`.
- Creation of an interceptor to extract user-id and user-role from request headers - `UserHeaderInterceptor`.
- Integration of Spotless with Google Java Formatter (AOSP) for consistent code formatting and linting.
- Setup of an environment for integration testing.
- Addition of **GitHub Actions** workflows for Clabbert to enable continuous integration and automated checks.

image
image

### DIFF
--- a/.github/workflows/clabbert.yml
+++ b/.github/workflows/clabbert.yml
@@ -1,0 +1,55 @@
+name: clabbert-ci
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "implemntation/clabbert/**"
+      - ".github/workflows/clabbert.yml"
+  pull_request:
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
+      - ready_for_review
+    branches:
+      - main
+    paths:
+      - "implemntation/clabbert/**"
+      - ".github/workflows/clabbert.yml"
+
+jobs:
+  build:
+    if: github.event.pull_request.draft == false
+    name: Build and test project
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    defaults:
+      run:
+        working-directory: implementation/clabbert
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup JDK
+        uses: actions/setup-java@v4
+        with:
+          java-version: "21"
+          distribution: "microsoft"
+
+      - name: Check formatting
+        run: ./gradlew spotlessCheck
+
+      - name: Run Build
+        run: ./gradlew assemble
+
+      - name: Run Unit Tests
+        run: ./gradlew test
+
+      - name: Run Integration Tests
+        run: ./gradlew integrationTest
--- a/.github/workflows/clabbert.yml
+++ b/.github/workflows/clabbert.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - main
     paths:
-      - "implemntation/clabbert/**"
+      - "implementation/clabbert/**"
       - ".github/workflows/clabbert.yml"
   pull_request:
     types:
@@ -17,7 +17,7 @@ on:
     branches:
       - main
     paths:
-      - "implemntation/clabbert/**"
+      - "implementation/clabbert/**"
       - ".github/workflows/clabbert.yml"
 
 jobs:
@@ -33,8 +33,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
 
       - name: Setup JDK
         uses: actions/setup-java@v4
--- a/implementation/clabbert/build.gradle
+++ b/implementation/clabbert/build.gradle
@@ -1,42 +1,62 @@
 plugins {
-	id 'java'
-	id 'org.springframework.boot' version '3.4.1'
-	id 'io.spring.dependency-management' version '1.1.7'
+    id 'java'
+    id 'org.springframework.boot' version '3.4.1'
+    id 'io.spring.dependency-management' version '1.1.7'
+    id("com.diffplug.spotless") version "6.25.0"
 }
 
 group = 'pwr.jakprzyjade'
 version = '0.0.1'
 
 java {
-	toolchain {
-		languageVersion = JavaLanguageVersion.of(21)
-	}
+    toolchain {
+        languageVersion = JavaLanguageVersion.of(21)
+    }
 }
 
 configurations {
-	compileOnly {
-		extendsFrom annotationProcessor
-	}
+    compileOnly {
+        extendsFrom annotationProcessor
+    }
 }
 
 repositories {
-	mavenCentral()
+    mavenCentral()
 }
 
 dependencies {
-	implementation 'org.springframework.boot:spring-boot-starter-actuator'
-	implementation 'org.springframework.boot:spring-boot-starter-data-jpa'
-	implementation 'org.springframework.boot:spring-boot-starter-web'
-	implementation 'org.springframework.boot:spring-boot-starter-validation'
-	implementation 'org.springdoc:springdoc-openapi-starter-webmvc-ui:2.7.0'
-	compileOnly 'org.projectlombok:lombok'
-	runtimeOnly 'org.postgresql:postgresql'
-	annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
-	annotationProcessor 'org.projectlombok:lombok'
-	testImplementation 'org.springframework.boot:spring-boot-starter-test'
-	testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
+    implementation 'org.springframework.boot:spring-boot-starter-actuator'
+    implementation 'org.springframework.boot:spring-boot-starter-data-jpa'
+    implementation 'org.springframework.boot:spring-boot-starter-web'
+    implementation 'org.springframework.boot:spring-boot-starter-validation'
+    implementation 'org.springdoc:springdoc-openapi-starter-webmvc-ui:2.7.0'
+    compileOnly 'org.projectlombok:lombok'
+    runtimeOnly 'org.postgresql:postgresql'
+    annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
+    annotationProcessor 'org.projectlombok:lombok'
+    testImplementation 'org.springframework.boot:spring-boot-starter-test'
+    testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
 }
 
 tasks.named('test') {
-	useJUnitPlatform()
+    useJUnitPlatform()
+}
+
+spotless {
+    format 'misc', {
+        target '*.gradle', '.gitattributes', '.gitignore'
+        trimTrailingWhitespace()
+        indentWithSpaces(4)
+        endWithNewline()
+    }
+    java {
+        googleJavaFormat('1.21.0').aosp().reflowLongStrings().skipJavadocFormatting()
+        formatAnnotations()
+
+        if (project.hasProperty('projSpotlessTarget')) {
+            target project.projSpotlessTarget
+        } else {
+            target 'src/*/java/**/*.java'
+        }
+    }
 }
--- a/implementation/clabbert/build.gradle
+++ b/implementation/clabbert/build.gradle
@@ -22,10 +22,6 @@ sourceSets {
         java {
             srcDirs 'src/integration/java'
         }
-
-        tasks.named('compileIntegrationJava', JavaCompile) {
-            options.compilerArgs += '-Xlint:-unused'
-        }
     }
 }
 
--- a/implementation/clabbert/build.gradle
+++ b/implementation/clabbert/build.gradle
@@ -14,10 +14,27 @@ java {
     }
 }
 
+sourceSets {
+    integration {
+        compileClasspath += sourceSets.main.output
+        runtimeClasspath += sourceSets.main.output
+
+        java {
+            srcDirs 'src/integration/java'
+        }
+
+        tasks.named('compileIntegrationJava', JavaCompile) {
+            options.compilerArgs += '-Xlint:-unused'
+        }
+    }
+}
+
 configurations {
     compileOnly {
         extendsFrom annotationProcessor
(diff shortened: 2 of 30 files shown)


implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/UserHeaderInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+            Arrays.stream(UserRole.values()).map(Enum::name).toList();
+
+    @Override
+    public boolean preHandle(
+            @NonNull HttpServletRequest request,
+            @NonNull HttpServletResponse response,
+            @NonNull Object handler)
+            throws Exception {
+        var userId = request.getHeader("user-id");
+        var userRole = request.getHeader("user-role");
```
jakubzehner: These are my proposed header names. What do you think of them?
@tchojnacki @mlodybercik @piterek130 

I also want to point out that, contrary to what @mlodybercik  said, requiring `X-` as the beginning of the header name is [deprecated](https://stackoverflow.com/a/3561399).

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/UserHeaderInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+
+        if (userId == null && userRoleEnum != UserRole.GUEST) {
+            throw new UserIdHeaderMissing();
+        }
+
+        if (requiredRoles.isPresent() && !List.of(requiredRoles.get()).contains(userRoleEnum)) {
+            throw new UserUnauthorized();
+        }
+
+        request.setAttribute("userId", userId);
```
jakubzehner: I will convert String to UUID later, in upcoming PRs

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/UserHeaderInterceptor.java
tchojnacki: Okay, it would hard to find a better source than RFC, so I guess the `X-` is out of the picture now. Good find 👍

.github/workflows/clabbert.yml
```diff
@@ -0,0 +1,55 @@
+name: clabbert-ci
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "implemntation/clabbert/**"
```
tchojnacki: **issue:** `implemntation`? - looks like a typo

.github/workflows/clabbert.yml
```diff
@@ -0,0 +1,55 @@
...
+    types:
+      - opened
+      - edited
+      - reopened
+      - synchronize
+      - ready_for_review
+    branches:
+      - main
+    paths:
+      - "implemntation/clabbert/**"
```
tchojnacki: **issue:** `implemntation`

.github/workflows/clabbert.yml
```diff
@@ -0,0 +1,55 @@
...
+    timeout-minutes: 10
+    defaults:
+      run:
+        working-directory: implementation/clabbert
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
tchojnacki: **question:** Why do we need earlier commits?

implementation/clabbert/build.gradle
```diff
@@ -1,42 +1,100 @@
 plugins {
-	id 'java'
-	id 'org.springframework.boot' version '3.4.1'
-	id 'io.spring.dependency-management' version '1.1.7'
+    id 'java'
+    id 'org.springframework.boot' version '3.4.1'
+    id 'io.spring.dependency-management' version '1.1.7'
+    id("com.diffplug.spotless") version "7.0.0.BETA4"
```
tchojnacki, line 5: **nit:** Should we use a beta version here instead of last stable?

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/configurations/ApiConfig.java
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.lang.NonNull;
+import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+import pwr.jakprzyjade.clabbert.api.middlewares.UserHeaderInterceptor;
+
+@Configuration
+@RequiredArgsConstructor
+public class ApiConfig implements WebMvcConfigurer {
+
+    @Autowired private final UserHeaderInterceptor userHeaderInterceptor;
```
tchojnacki: **suggestion:** It is commonly believed in that in newer Spring versions, constructor injection should be preferred to field injection:
- https://odrotbohm.de/2013/11/why-field-injection-is-evil/
- https://stackoverflow.com/questions/40620000/spring-autowired-on-properties-vs-constructor
- https://reflectoring.io/constructor-injection/

You even used `@RequiredArgsConstructor`, is there any reason to add the `@Autowired`?

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/ApiExceptionHandler.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @ExceptionHandler(AbstractException.class)
+    public ResponseEntity<ErrorDto> handleAbstractException(
+            AbstractException exception, WebRequest request) {
+        var kind =
+                exception
+                        .getClass()
+                        .getSimpleName()
+                        .replaceAll("([a-z])([A-Z])", "$1-$2")
+                        .toLowerCase();
```
tchojnacki: **thought:** Scary. Wouldn't it also keep the `Exception` suffix? I prefered `account-not-found` to `account-not-found-exception` since the last part is redundant, but I'm kinda fine with both.

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/ApiExceptionHandler.java
```diff
@@ -0,0 +1,63 @@
...
+
+        return new ResponseEntity<>(body, HttpStatus.NOT_FOUND);
+    }
+
+    @ExceptionHandler(Exception.class)
+    public ResponseEntity<ErrorDto> handleAll(Exception exception, WebRequest request) {
+        var body =
+                ErrorDto.builder()
+                        .code(HttpStatus.INTERNAL_SERVER_ERROR.value())
+                        .kind("internal-server-error")
```
tchojnacki:
```suggestion
                        .kind("server-failure")
```

**nit:** Can we change it like so to make it consistent with my version? I wanted to keep the error names separate from HTTP names (since errors are a part of the domain layer, and HTTP is in the presentation layer, so the errors should be independent).

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/UserHeaderInterceptor.java
tchojnacki: If we omit the `X-`, how about a different prefix to differentiate between our headers and standard ones? Like `jp-user-id`/`jp-user-role` (**J**ak**P**rzyjade). It's weird that `user-role` is controlled by use while `user-agent` isn't.

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/UserHeaderInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public boolean preHandle(
+            @NonNull HttpServletRequest request,
+            @NonNull HttpServletResponse response,
+            @NonNull Object handler)
+            throws Exception {
+        var userId = request.getHeader("user-id");
+        var userRole = request.getHeader("user-role");
+
```
tchojnacki, line 40: **suggestion:** Not sure if this is the right place to discuss, but I've also though about `request-id`, which would be generated by Phoenix as a random UUIDv7 and then passed to each subsequent service (and also passed along when passing other services through internal channels), this would make debugging less scary, since we can trace how requests go through different services.

https://microservices.io/patterns/observability/distributed-tracing.html

obraz

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/UserHeaderInterceptor.java
tchojnacki: We also need more conventions regarding, for instance is guest expressed as `user-role: null` or `"guest"`? What is the value of `user-id` for guests? `null`? empty? or no header at all? I'm fine with documenting it in Markdown myself, but there are still many undecided factors.

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/UserHeaderInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+        if (requiredRoles.isPresent() && !List.of(requiredRoles.get()).contains(userRoleEnum)) {
+            throw new UserUnauthorized();
+        }
+
+        request.setAttribute("userId", userId);
+        request.setAttribute("userRole", userRoleEnum);
+        return true;
+    }
+
+    private Optional<UserRole[]> getRequiredRoles(Object handler) {
```
tchojnacki: **nit:** Probably needs `@NonNull` somewhere, since `Optional` can be `empty`, `of`, but also `null` (I love Java). Looking now, wouldn't most of your methods require `@NonNull` return type annotation?

.github/workflows/clabbert.yml
jakubzehner: My bad, this is the result of not understanding and misinterpreting what fellow copilot has produced.

implementation/clabbert/build.gradle
jakubzehner: It's possible, but it's just a formatter, so I'll ignore your comment (especially since in the “stable” version for some reason putting the year in copyright comment doesn't work properly)

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/configurations/ApiConfig.java
jakubzehner: Full agreement, that's what I wanted to do too, `@Autowired` stayed there by accident. Thanks for catching that 👍

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/ApiExceptionHandler.java
jakubzehner: Correct, but there will be no `-exception` if there is no word “Exception” in the exception name. 
image


Although, in retrospect, they should include it in the name, and I'll change it so that it removes it from the exception name when converting to `kind`.

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/ApiExceptionHandler.java
jakubzehner: Sure, I'll change it

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/UserHeaderInterceptor.java
jakubzehner: I am then changing it to:
`jp-user-id`
`jp-user-role`
As discussed.

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/UserHeaderInterceptor.java
jakubzehner: I am then changing it to:
`jp-user-id`
`jp-user-role`
As discussed.

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/UserHeaderInterceptor.java
jakubzehner: We can do it, however, this is probably not the right interceptor for it. I will think about what to do with it when (if xD) I introduce logging.

implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/middlewares/UserHeaderInterceptor.java
jakubzehner: On the one hand this is true, on the other hand it seems to me that the minimum responsibility should be on the developer's side and if the Optional type is used then I assume in advance that there will never be a situation where null is returned.
Especially since even the description of this class in Java has such a piece of description:
> API Note:
>Optional is primarily intended for use as a method return type where there is a clear need to represent "no result," and where using null is likely to cause errors. A variable whose type is Optional should never itself be null; it should always point to an Optional instance.

In addition, I have no intention of placing the `@NonNull` annotation everywhere until the static analyzer tells me that it is expected in a given place.

tchojnacki:
Seems fine, I'll do another pass in the morning when my mind will be a bit fresher and probably (95%) approve it.

tchojnacki (approved):

PR #48, opened 2024-12-29
PR: `Leprechaun` line, stop, vehicle implementation.

mlodybercik:
obraz1
obraz2


implementation/leprechaun/.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
+    "recommendations": ["esbenp.prettier-vscode"]
```
tchojnacki: **nit:** Don't we also need ESLint or something? It seems to be used in `settings.json`.

implementation/leprechaun/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "@mikro-orm/nestjs": "^6.0.2",
+    "@mikro-orm/postgresql": "^6.4.2",
+    "@nestjs/common": "^10.0.0",
+    "@nestjs/core": "^10.0.0",
+    "@nestjs/platform-express": "^10.0.0",
+    "@nestjs/swagger": "^8.1.0",
+    "class-validator": "^0.14.1",
+    "jest-mock": "^29.7.0",
+    "reflect-metadata": "^0.2.0",
+    "rxjs": "^7.8.1",
```
tchojnacki, line 38: **nit:** Is `rxjs` used anywhere? 🤔

implementation/leprechaun/src/config/mikro-orm.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from '@mikro-orm/postgresql';
+
+export default defineConfig({
+  dbName: 'leprechaun',
+
+  entities: ['./dist/**/database/*.entity.js'],
+  entitiesTs: ['./src/**/database/*.entity.ts'],
+  clientUrl: 'postgresql://postgres:test@localhost:5432/leprechaun',
+  debug: true,
```
tchojnacki: **nit:** These two should probably be controlled by env variables, but we don't have them standardized yet, so it's fine.

implementation/leprechaun/src/internal/healthcheck/healthcheck.controller.ts
```diff
@@ -0,0 +1,15 @@
+import { Controller, Get } from '@nestjs/common';
+import { ApiInternalServerErrorResponse, ApiOkResponse } from '@nestjs/swagger';
+import { HealthcheckDto, HealthcheckStatus } from './healthcheck.dto';
+
+@Controller('/int/v1/health')
+export class HealthcheckController {
+  @Get('/')
+  @ApiOkResponse({ type: HealthcheckDto, description: 'Healthcheck status' })
+  @ApiInternalServerErrorResponse({ description: 'Internal server error' })
```
tchojnacki: > The endpoint MUST return a `200 OK` if the service is possibly healthy (`UP`, `UNKNOWN`), and a `503 Service Unavailable` if the service is unhealthy (`DOWN`, `OUT_OF_SERVICE`). [^1]

**issue:** It seems that you return `500` instead of `503`?

[^1]: https://github.com/PWR-ACS-SE-24/SoftwareSystemDesign/blob/main/documentation/contracts.md

implementation/leprechaun/src/internal/healthcheck/healthcheck.dto.ts
```diff
@@ -0,0 +1,14 @@
+import { ApiProperty } from '@nestjs/swagger';
+
+export enum HealthcheckStatus {
+  UP = 'UP',
+  DOWN = 'DOWN',
+  UNKNOWN = 'UNKNOWN',
+}
```
tchojnacki: **thought:** This can stay as-is, I was just kind of surprised, since it's probably the first time in 2-3 years where I've seen an `enum` in TypeScript.

- https://www.totaltypescript.com/why-i-dont-like-typescript-enums
- https://www.youtube.com/watch?v=jjMbPt_H3RQ
- https://frontlive.pl/blog/typescript-enum
- https://stackoverflow.com/q/40275832
- https://www.typescriptcourse.com/string-literal-unions-over-enums

implementation/leprechaun/src/internal/internal.module.ts
```diff
@@ -0,0 +1,8 @@
+import { Module } from '@nestjs/common';
+import { HealthcheckController } from './healthcheck/healthcheck.controller';
+import { RoutesController } from './routes/routes.controller';
+
+@Module({
+  controllers: [HealthcheckController, RoutesController]
```
tchojnacki: **nit:** Do you really need two one-endpoint controllers instead of defining something like `MonitoringController` with two routes?

implementation/leprechaun/src/main.ts
```diff
@@ -0,0 +1,25 @@
...
+  const config = new DocumentBuilder()
+    .setTitle('Leprechaun API')
+    .setDescription('API for the Leprechaun service')
+    .setVersion('1.0')
+    .build();
+
+  const documentFactory = () => SwaggerModule.createDocument(app, config);
+  SwaggerModule.setup('/int/v1/docs/swagger', app, documentFactory, {
+    jsonDocumentUrl: '/int/v1/docs/openapi.json',
+  });
```
tchojnacki: **issue:** Having one shared REST API between internal and external routes is problematic:

1. If we ever decide to create an OpenAPI spec for API Gateway by merging the OpenAPIs for services, we would leak internal endpoints.
2. Internal and external endpoint might (and ideally would) have different interceptors, validation, etc. (e.g. external endpoints need `jp-user-id` and `jp-user-role` headers, while internal ones don't; external endpoints might be rate limited, while internal shouldn't; external endpoints should be more aggressively cached).
3. We are mixing two different interfaces in one port, if we decided to switch to GraphQL for external endpoints or to Cap’n Proto for internal endpoints the current approach would break.
4. We might want to host internal and external endpoints at different ports for increased security.

I could make up more practical arguments. I am not heavily advocating for the change, since our system should be a proof of concept instead of an industrial solution, if you feel like this is sufficient let it be.

As for theoretical arguments instead of practical ones, this speaks to me:

Untitled Diagram drawio

implementation/leprechaun/src/shared/api/generic-paginated.dto.ts
```diff
@@ -0,0 +1,30 @@
...
+  static fromEntities<TDto>(total: number, size: number, page: number, entities: TDto[]): PaginatedDto<TDto> {
+    return new PaginatedDto<TDto>(total, size, page, entities);
+  }
+
+  static sanitizePagination(size: number, page: number): { size: number; page: number } {
+    return {
+      page: Math.max(0, page),
+      size: Math.min(100, Math.max(1, size)),
+    };
+  }
```
tchojnacki: **nit:** Would this be the responsibility of the DTO? Or rather the application layer.

implementation/leprechaun/src/shared/api/generic-paginated.dto.ts
```diff
@@ -0,0 +1,30 @@
...
+  readonly page: number;
+
+  constructor(total, size, page, data) {
+    this.total = total;
+    this.size = size;
+    this.page = page;
+    this.data = data;
+  }
+
+  static fromEntities<TDto>(total: number, size: number, page: number, entities: TDto[]): PaginatedDto<TDto> {
```
tchojnacki: **very very nit:** You are calling this method `fromEntities` and it takes DTOs instead of entities as arguments.

implementation/leprechaun/src/shared/api/http-exceptions.ts
```diff
@@ -0,0 +1,69 @@
...
+  @ApiProperty({ description: 'Error message in Polish', required: false })
+  messagePl?: string;
+
+  constructor(kind: ErrorKind, code: number, messageEn: string, messagePl: string) {
+    this.kind = kind;
+    this.code = code;
+    this.messageEn = messageEn;
+    this.messagePl = messagePl;
+  }
+}
```
tchojnacki: **nit:** The current approach lets you create errors with mismatched `kind` and `code`, which can be prevented with TS's type system.

implementation/leprechaun/src/shared/api/http-exceptions.ts
```diff
@@ -0,0 +1,69 @@
...
+// Typical HTTP exceptions
+export const exceptionMap = {
+  SchemaMismatchException: () =>
+    new HttpExceptionDto(
+      ErrorKind.SchemaMismatchException,
+      HttpStatus.UNPROCESSABLE_ENTITY,
+      'Schema mismatch exception',
+      'Niezgodność schematu',
+    ),
+  NotFoundException: (id: any) => {
```
tchojnacki:
```suggestion
  NotFoundException: (id: unknown) => {
```

implementation/leprechaun/src/shared/api/http-exceptions.ts
```diff
@@ -0,0 +1,69 @@
...
+      HttpStatus.INTERNAL_SERVER_ERROR,
+      'Internal server error',
+      'Wewnętrzny błąd serwera',
+    ),
+};
+
+export function mapException<T extends HttpException>(error: T): AppError {
+  try {
+    return exceptionMap[error.constructor.name](error.getResponse());
+  } catch (e) {
```
tchojnacki:
```suggestion
  } catch {
```

implementation/leprechaun/src/shared/api/generic-paginated.ts
```diff
@@ -0,0 +1,26 @@
+import { applyDecorators, Type } from '@nestjs/common';
+import { ApiExtraModels, ApiOkResponse, ApiQuery, getSchemaPath } from '@nestjs/swagger';
+import { PaginatedDto } from './generic-paginated.dto';
+
+export const ApiPaginatedResponse = <TModel extends Type<any>>(model: TModel) => {
```
tchojnacki:
```suggestion
export const ApiPaginatedResponse = <TModel extends Type>(model: TModel) => {
```

implementation/leprechaun/src/shared/database/uuid.ts
tchojnacki, line 1: **nit:** Shouldn't this belong to application layer instead of DB?

implementation/leprechaun/src/shared/api/validation.service.ts
```diff
@@ -0,0 +1,15 @@
+import { Injectable } from '@nestjs/common';
+import { validate } from 'class-validator';
+import { SchemaMismatchException } from './http-exceptions';
+
+@Injectable()
+export class ValidationService {
+  // We could possible redo this as validation-pipe, CreateObjectValidationPipe and PatchObjectValidationPipe
+  async validate<T extends object>(obj: T, update: boolean = false): Promise<boolean> {
```
tchojnacki: **question:** Is there any situation where this would return a `false`? If not, the signature can be changed to `Promise<true>`, and if that's the case, it might as well be `Promise<void>` since the result is useless.

implementation/leprechaun/src/vehicle/controller/vehicle.controller.ts
```diff
@@ -0,0 +1,75 @@
...
+import { PaginatedDto } from '../../shared/api/generic-paginated.dto';
+import { HttpExceptionDto, SchemaMismatchException } from '../../shared/api/http-exceptions';
+import { VehicleService } from '../service/vehicle.service';
+import { CreateVehicleDto, UpdateVehicleDto } from './vehicle-create.dto';
+import { VehicleDto } from './vehicle.dto';
+
+export const UUIDPipe = new ParseUUIDPipe({
+  version: '7',
+  exceptionFactory: () => new SchemaMismatchException(),
+});
```
tchojnacki: **question:** Does this allow nil UUIDs? And should this be defined here, if we have UUIDs all around?

implementation/leprechaun/src/vehicle/controller/vehicle.controller.ts
```diff
@@ -0,0 +1,75 @@
...
+
+@Controller('/ext/v1/vehicles')
+@ApiExtraModels(PaginatedDto)
+export class VehicleController {
```
tchojnacki, line 20: LGTM 👍

implementation/leprechaun/src/vehicle/controller/vehicle.dto.ts
```diff
@@ -0,0 +1,43 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { Vehicle } from '../database/vehicle.entity';
+
+export class VehicleDto {
+  @ApiProperty({
+    description: 'Vehicle ID',
+    maxLength: 36,
+    minLength: 36,
+    nullable: false,
+    examples: ['f1b1b9b1-1c1b-4b1b-9b1b-1c1b1b1b1b1b'],
```
tchojnacki, line 12: **question:** Does `format: "uuid"` work here?

implementation/leprechaun/src/vehicle/service/vehicle.service.ts
```diff
@@ -0,0 +1,53 @@
+import { EntityRepository } from '@mikro-orm/core';
+import { InjectRepository } from '@mikro-orm/nestjs';
+import { Injectable, NotFoundException } from '@nestjs/common';
+import { ValidationService } from '../../shared/api/validation.service';
+import { CreateVehicleDto, UpdateVehicleDto } from '../controller/vehicle-create.dto';
+import { Vehicle } from '../database/vehicle.entity';
+
+@Injectable()
+export class VehicleService {
```
tchojnacki, line 9: LGTM 👍

implementation/leprechaun/src/vehicle/test/vehicle.controller.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+            updateVehicleById: jest.fn().mockResolvedValue(VEHICLES[0]),
+          };
+          return mocked;
+        }
+        if (typeof token === 'function') {
+          const mockMetadata = moduleMocker.getMetadata(token) as MockFunctionMetadata<any, any>;
+          const Mock = moduleMocker.generateFromMetadata(mockMetadata);
+          return new Mock();
+        }
+      })
```
tchojnacki: I hate JavaScript, LGTM though 👍

implementation/leprechaun/src/vehicle/test/vehicle.controller.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+              VEHICLES.push(vehicle);
+              return vehicle;
+            }),
+            deleteVehicleById: jest.fn().mockResolvedValue(undefined),
+            updateVehicleById: jest.fn().mockResolvedValue(VEHICLES[0]),
+          };
+          return mocked;
+        }
+        if (typeof token === 'function') {
+          const mockMetadata = moduleMocker.getMetadata(token) as MockFunctionMetadata<any, any>;
```
tchojnacki: `<any, any>` 😠

implementation/leprechaun/tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
+    "allowSyntheticDefaultImports": true,
+    "target": "ES2021",
+    "sourceMap": true,
+    "outDir": "./dist",
+    "baseUrl": "./",
+    "incremental": true,
+    "skipLibCheck": true,
+    "strictNullChecks": false,
```
tchojnacki:
```suggestion
    "strictNullChecks": true,
```

We already have Java in the repository, let's not have its clone in JavaScript.

`const test: string = null;` passes with this set to `false` and doesn't with it set to `true`. To be honest, you should probably enable `strict` https://www.typescriptlang.org/tsconfig/#strict.

tchojnacki (commented):
**nit:** Why are there so many missing newlines at the end of the file? It breaks the subsequent diffs, makes the lines possibly not get counted by stuff like `wc` and violates POSIX standards [[1]](https://stackoverflow.com/q/729692) [[2]](https://stackoverflow.com/q/5813311).

**issue:** The screenshot in the PR seems to be outdated? Is healthcheck still at `/healthcheck`?

implementation/leprechaun/package.json
mlodybercik: `rxjs` is used internally by `nestjs` and by custom interceptors outside internals. I'll probably use them (?). Thus leaving it for now.

See [here](https://docs.nestjs.com/interceptors#aspect-interception).

implementation/leprechaun/src/config/mikro-orm.config.ts
mlodybercik: Yeah, when I'll have some code that does more than plain CRUD, I'll make it configurable from outside. For now pointing to localhost `as const`.

implementation/leprechaun/src/internal/healthcheck/healthcheck.controller.ts
mlodybercik: Right! My bad.

implementation/leprechaun/src/internal/healthcheck/healthcheck.dto.ts
mlodybercik: OpenAPI supports them nicely, and I don't believe that I'll be doing anything fancy with them. I'll change them just in case.

implementation/leprechaun/src/internal/internal.module.ts
mlodybercik: I know already that I will have to change package diagram from E2, so, fair. D:

implementation/leprechaun/src/main.ts
mlodybercik: That was on my todo list. I know that `nestjs` can output two different openapi swaggers and even intercept requests to them. I'll do that now.

implementation/leprechaun/src/config/mikro-orm.config.ts
jakubzehner: Well... I would not standarize it xD
https://github.com/PWR-ACS-SE-24/SoftwareSystemDesign/blob/main/implementation%2Fclabbert%2Fcompose.yml#L28-L30

implementation/leprechaun/tsconfig.json
jakubzehner: Oh wow, no need to be offensive xD. Let's not start a classic war about languages. Especially since neither Java nor Js is in the best position to argue about type safety

implementation/leprechaun/src/shared/api/http-exceptions.ts
mlodybercik: I'll turn this enum into object the same way I did for Healthcheck, and I'll keep `(kind, status_code)` as a tuple for creating DTOs.

implementation/leprechaun/src/shared/api/generic-paginated.ts
mlodybercik: I did `<any>` because it was previously complaining about types, but now it doesn't? huh. Sometimes my vscode does weird things and I have to reload the window for it to work.

implementation/leprechaun/src/shared/database/uuid.ts
mlodybercik: I want to make a `MicroORM`s type for dealing with UUID's, as currently UUID are passed around as `string`s.

implementation/leprechaun/src/vehicle/controller/vehicle.controller.ts
mlodybercik: 1) It does not. Implementation checks whether it was set as optional (in this case it is not, as its set by default to false) and then does regex match against proper uuid version.
2) I moved it to a different file in `shared` module, as its needed by all modules.

implementation/leprechaun/src/vehicle/controller/vehicle.dto.ts
mlodybercik:
```ts
@ApiProperty({
  format: 'uuid',
  description: 'Vehicle ID',
  nullable: false,
})
```
obraz

```ts
@ApiProperty({
  description: 'Vehicle ID',
  maxLength: 36,
  minLength: 36,
  nullable: false,
  examples: ['f1b1b9b1-1c1b-4b1b-9b1b-1c1b1b1b1b1b'],
})
```
obraz

Neither of them look proper to me.

implementation/leprechaun/src/vehicle/test/vehicle.controller.spec.ts
mlodybercik: It's just mocks and `jest`, nothing to see here anyway.

implementation/leprechaun/tsconfig.json
mlodybercik: I did
```jsonc
// ...
    "strict": true,
    "strictNullChecks": true,
    "noImplicitAny": true,
    "strictBindCallApply": true,
// ...
```
let's see how long till depression kicks in.

mlodybercik: nvm, i've started hearing voices. I'll make only `strict` true.

implementation/leprechaun/src/vehicle/test/vehicle.controller.spec.ts
mlodybercik: I have no idea what this piece of code even does. And it doesn't break any tests when I remove it!

implementation/leprechaun/.vscode/extensions.json
mlodybercik: I find ESLint more of a hurdle than a tool that helps. It checks for the exact same things as linter in vscode does.

implementation/leprechaun/src/shared/api/generic-paginated.dto.ts
mlodybercik: Yeah, naming problem.

implementation/leprechaun/src/shared/api/generic-paginated.dto.ts
mlodybercik: It is, but I keep it here because whenever I use `PaginatedDto`, I have to sanitize parameters, so it is a little bit cleaner than creating new file with 5 lines. I maybe will in the future make a `Pipe` for these.

implementation/leprechaun/src/internal/controller/healthcheck.dto.ts
```diff
@@ -0,0 +1,17 @@
...
+} as const;
+
+type Values<Object> = Object[keyof Object];
+
+export class HealthcheckDto {
+  @ApiProperty({ enum: () => Object.keys(HealthcheckStatus), example: HealthcheckStatus.UP })
+  status: Values<typeof HealthcheckStatus>;
+
+  // TODO: add more fields
+}
```
tchojnacki: Huh? That's somehow worse XD

```suggestion
export type HealthcheckStatus = 'UP' | 'DOWN' | 'UNKNOWN' | 'OUT_OF_SERVICE';

export class HealthcheckDto {
  @ApiProperty({ enum: ['UP', 'DOWN', 'UNKNOWN', 'OUT_OF_SERVICE'], example: 'UP' })
  status: HealthcheckStatus;

  // TODO: add more fields
}
```

implementation/leprechaun/src/internal/controller/monitoring.controller.ts
```diff
@@ -0,0 +1,26 @@
...
+  @ApiOkResponse({ type: HealthcheckDto, description: 'Healthcheck status' })
+  @ApiResponse({
+    type: HealthcheckDto,
+    description: 'Internal server error',
+    status: HttpStatus.SERVICE_UNAVAILABLE,
+    example: { status: HealthcheckStatus.OUT_OF_SERVICE },
+  })
+  async healthcheck(): Promise<HealthcheckDto> {
+    return {
+      status: HealthcheckStatus.UP,
```
tchojnacki:
```suggestion
    example: { status: 'OUT_OF_SERVICE' },
  })
  async healthcheck(): Promise<HealthcheckDto> {
    return {
      status: 'UP',
```

implementation/leprechaun/.vscode/extensions.json
tchojnacki: I agree, but can you do `"source.fixAll.eslint": "always"` (present in `settings.json`) without any additional extension? 🤔

implementation/leprechaun/src/config/mikro-orm.config.ts
tchojnacki: > Well... I would not standarize it xD https://github.com/PWR-ACS-SE-24/SoftwareSystemDesign/blob/main/implementation%2Fclabbert%2Fcompose.yml#L28-L30

Great, I'm gathering such comments for the inevitable "why did you guys decide to not write code in Spring" discussion 👍

implementation/leprechaun/src/vehicle/controller/vehicle.dto.ts
tchojnacki: I'd combine `format: "uuid"` and your own `examples`.

implementation/leprechaun/src/vehicle/test/vehicle.controller.spec.ts
tchojnacki: Great 👌

implementation/leprechaun/tsconfig.json
tchojnacki: > I'll make only strict true.

Strict enables many other settings and I'd say that it is probably enough for your needs.

implementation/leprechaun/src/main.ts
```diff
@@ -0,0 +1,51 @@
...
+    .setTitle('Leprechaun API')
+    .setDescription('API for the Leprechaun service')
+    .setVersion('1.0')
+    .build();
+
+  const externalConfig = new DocumentBuilder()
+    .setTitle('Leprechaun API')
+    .setDescription('API for the Leprechaun service')
+    .setVersion('1.0')
+    .build();
```
tchojnacki: **nit:** I'd add something in their name (like "internal"/"external") to make them easier to differentiate, but I guess we can infer that from the URL.

tchojnacki (commented):
I'll go over the whole thing top to bottom in a few minutes since this PR is huge and I'm kind of getting lost. Seems fine though.

implementation/leprechaun/.prettierrc
tchojnacki, line 1: **nit:** This can be renamed to `.prettierrc.json` to enable syntax highlighting, formatting, etc. without changes to functionality.

implementation/leprechaun/src/line/test/line.controller.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+import { LineController } from '../controller/line.controller';
+import { Line } from '../database/line.entity';
+import { LineService } from '../service/line.service';
+
+const moduleMocker = new ModuleMocker(global);
+
+const UUID_SAMPLE = '019411be-c6c3-7f67-83c1-c088d4280102';
+const LINE: Line = {
+  id: UUID_SAMPLE,
+  name: 'Klinii',
```
tchojnacki: **very nit:** Weird line name.

implementation/leprechaun/src/shared/api/http-exceptions.ts
```diff
@@ -0,0 +1,62 @@
+import { HttpException, HttpStatus } from '@nestjs/common';
+import { ApiProperty } from '@nestjs/swagger';
+
+export const ErrorKind: Record<string, [string, number]> = {
+  SchemaMismatchException: ['schema-mismatch-exception', HttpStatus.UNPROCESSABLE_ENTITY],
+  NotFoundException: ['resource-not-found-exception', HttpStatus.NOT_FOUND],
+  InternalServerError: ['internal-server-error', HttpStatus.INTERNAL_SERVER_ERROR],
+} as const;
```
tchojnacki: **suggestion:** The type annotation discards `as const`. You probably wanted to do this:
```suggestion
export const ErrorKind = {
  SchemaMismatchException: ['schema-mismatch-exception', HttpStatus.UNPROCESSABLE_ENTITY],
  NotFoundException: ['resource-not-found-exception', HttpStatus.NOT_FOUND],
  InternalServerError: ['internal-server-error', HttpStatus.INTERNAL_SERVER_ERROR],
} as const satisfies Record<string, [string, number]>;
```

implementation/leprechaun/src/internal/controller/healthcheck.dto.ts
mlodybercik: Huh, that was in one of the examples you gave xD

implementation/leprechaun/src/shared/api/http-exceptions.ts
```diff
@@ -0,0 +1,62 @@
...
+
+interface AppError {
+  kind: string;
+  code: number;
+  messageEn: string;
+  messagePl?: string;
+}
+
+export class HttpExceptionDto implements AppError {
+  @ApiProperty({ description: 'Error kind', enum: () => Object.values(ErrorKind), isArray: false })
```
tchojnacki: **suggestion:** `enum` can take an array as value, is there any reason to do
```
() => Object.values(ErrorKind)
```
instead of
```
Object.values(ErrorKind)
```
?

(applies to multiple places, but I'll only leave a comment here)

implementation/leprechaun/src/shared/api/http-exceptions.ts
tchojnacki:
```TS
new HttpExceptionDto(['schema-mismatch-exception', HttpStatus.INTERNAL_SERVER_ERROR], '', '');
```

Passes your version and fails on mine with `Type 'HttpStatus.INTERNAL_SERVER_ERROR' is not assignable to type 'HttpStatus.UNPROCESSABLE_ENTITY'.`

tchojnacki (approved):

implementation/leprechaun/src/config/mikro-orm.config.ts
jakubzehner: well, okay, fine, we can standardize, but let's break it down into ip, port, db name, user, password instead of one url

mlodybercik: I just need a list of variables that have to be present, thats all.

<details>
  <sup>any self-respecting framework uses DB_URI as a single variable...</sup>
 </details>

mlodybercik:
I know that I've changed again most of the files in this PR, after it has been fixed. But before it was even accepted I've managed to find some issues with my assumptions and had to change some things. Namely:
- tests sucked: mocking was f*cking awful, no db for ORM tests.
- `ManyToMany` in Mikroorm are built on `Set`s, so it was a `OrderedSet` by default. Fixed by changing to two `OneToMany` relations.

implementation/leprechaun/src/internal/controller/monitoring.controller.ts
```diff
@@ -0,0 +1,26 @@
...
+
+@Controller('/int/v1/')
+export class MonitoringController {
+  @Get('/health')
+  @ApiOkResponse({ type: HealthcheckDto, description: 'Healthcheck status' })
+  @ApiResponse({
+    type: HealthcheckDto,
+    description: 'Internal server error',
+    status: HttpStatus.SERVICE_UNAVAILABLE,
+    example: { status: <HealthcheckStatus>'UP' },
```
tchojnacki:
```suggestion
    example: { status: 'UP' },
```

implementation/leprechaun/src/line/controller/line-create.dto.ts
```diff
@@ -0,0 +1,45 @@
...
+export class CreateLineDto {
+  @IsString()
+  @IsNotEmpty()
+  @MaxLength(4)
+  @ApiProperty({
+    description: 'Name of the line',
+    maxLength: 4,
+    minLength: 1,
+    nullable: false,
+    examples: ['112', '997', '144', '145'],
```
tchojnacki: **nit:** Might add a line using alphabetic characters to highlight that such thing is allowed, eg `N` or `11A`.

implementation/leprechaun/src/line/controller/line-create.dto.ts
```diff
@@ -0,0 +1,45 @@
...
+  })
+  readonly name: string;
+
+  @IsOptional()
+  @IsArray({ always: true })
```
tchojnacki: **question:** Why `{ always: true }`? You don't seem to be using groups.

implementation/leprechaun/src/line/controller/line.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+@ApiExtraModels(StopDto)
+@ApiExtraModels(PaginatedDto)
+export class LineController {
+  constructor(private readonly lineService: LineService) {}
+
+  @Get('/')
+  @ApiPaginatedResponse(LineDto)
+  async getAllLines(
+    @Query('page') page_: number = 0,
+    @Query('size') size_: number = 50,
```
tchojnacki: obraz
```suggestion
    @Query('size') size_: number = 10,
```

implementation/leprechaun/src/line/controller/line.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+    // @Query('filter') TODO: add filter
+  ): Promise<PaginatedDto<LineDto>> {
+    const { size, page } = PaginatedDto.sanitizePagination(size_, page_);
```
tchojnacki: **suggestion:** With the current design you will have to duplicate this code across, making sure to not break anything (you can just forget `sanitizePagination` and it would still work, yet incorrectly), making sure all of the defaults are perserved etc.

For instance, per my last comment, you have an incorrect default `size`, meaning you have to go around and manually change all paginated endpoints.

You can instead create a custom query decorator to extract pagination query, here's an example: [definition](https://github.com/CodersCamp2021-HK/CodersCamp2021.Project.Fullstack/blob/main/packages/server/src/shared/api/decorators/PaginationDecorator.ts#L7), [usage](https://github.com/CodersCamp2021-HK/CodersCamp2021.Project.Fullstack/blob/main/packages/server/src/restaurants/dishes/api/RestaurantDishController.ts#L30). This would provide a shared place to define A) query parameter names, B) sanitization, C) defaults, so that they are not copied around.

---

The `sanitizePagination` function is generally a code smell, see these resources for an explaination:
- https://fsharpforfunandprofit.com/posts/designing-with-types-making-illegal-states-unrepresentable/
- https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/
- https://v5.chriskrycho.com/journal/making-illegal-states-unrepresentable-in-ts/
- https://ybogomolov.me/making-illegal-states-unrepresentable

implementation/leprechaun/src/line/controller/line-create.dto.ts
```diff
@@ -0,0 +1,45 @@
+import { ApiProperty, PartialType } from '@nestjs/swagger';
+import { IsArray, IsNotEmpty, IsOptional, IsString, IsUUID, MaxLength } from 'class-validator';
+
+export class CreateLineDto {
```
tchojnacki: **suggestion:** Can't `CreateLineDto` be created using `LineDto` and [`OmitType`](https://docs.nestjs.com/openapi/mapped-types#omit)?

implementation/leprechaun/src/line/database/stop-line-mapping.entity.ts
```diff
@@ -0,0 +1,27 @@
...
+  @ManyToOne({ primary: false, entity: () => Line })
+  line: Line;
+
+  @ManyToOne({ primary: false, entity: () => Stop })
+  stop: Stop;
+
+  @Property({ nullable: false, type: 'integer' })
+  order: number;
+
+  [PrimaryKeyProp]?: ['id'];
```
tchojnacki, line 20: I see that MikroORM is a great library 👁️👄👁️

implementation/leprechaun/src/line/service/line.service.ts
```diff
@@ -0,0 +1,96 @@
...
+    });
+    const total = await this.lineRepository.count();
+
+    return { lines, total };
+  }
+
+  async getLineById(lineId: string, filters: boolean = true): Promise<Line> {
+    const line = await this.lineRepository.findOneOrFail(
+      { id: lineId },
+      { failHandler: () => new NotFoundException(lineId), populate: ['mappings.stop'], filters: filters },
```
tchojnacki:
```suggestion
      { failHandler: () => new NotFoundException(lineId), populate: ['mappings.stop'], filters },
```

implementation/leprechaun/src/line/service/line.service.ts
```diff
@@ -0,0 +1,96 @@
...
+    ).map((mapping) => mapping.line);
+  }
+
+  async createLine(lineCreateDto: CreateLineDto): Promise<Line> {
+    return await this.em.transactional(async () => {
+      const line = new Line(lineCreateDto.name);
+      await this.lineRepository.insert(line);
+
+      // If lines were defined, check if they exist
+      if (lineCreateDto.stops) {
```
tchojnacki, line 60: **question:** Why is this conditional necessary? `stops` is a non-nullable, non-undefined array and I don't think they can ever coerce to `false`, even `Boolean([]) === true`.

implementation/leprechaun/src/line/service/line.service.ts
```diff
@@ -0,0 +1,96 @@
...
+    // (as discussed with @tchojnacki)
+    // simple, just mark as inactive and create a new one
+    const line = await this.getLineById(lineId);
+    return await this.em.transactional(async () => {
+      await this.lineRepository.nativeUpdate({ id: lineId }, { isActive: false });
+
+      return await this.createLine({
+        ...line,
+        ...lineUpdateDto,
+        stops: lineUpdateDto.stops || line.mappings.map((i) => i.id),
```
tchojnacki, line 86:
```suggestion
        stops: lineUpdateDto.stops ?? line.mappings.map((i) => i.id),
```

implementation/leprechaun/src/main.ts
```diff
@@ -0,0 +1,54 @@
...
+import { InternalModule } from './internal/internal.module';
+import { LineModule } from './line/line.module';
+import { RouteModule } from './route/route.module';
+import { HttpExceptionFilter, InternalExceptionFilter } from './shared/api/http-exception.filter';
+import { StopModule } from './stop/stop.module';
+import { VehicleModule } from './vehicle/vehicle.module';
+
+async function bootstrap() {
+  // TODO: Most DTOs contain repeating fields like `id`, `isActive`. Create a base class that contains these fields and extend it in all DTOs.
+  // TODO: We propably need explicit transactins in some places when doing more than single operation.
```
tchojnacki, line 15:
```suggestion
  // TODO: We probably need explicit transactions in some places when doing more than a single operation.
```

implementation/leprechaun/src/main.ts
```diff
@@ -0,0 +1,54 @@
...
+      include: [LineModule, StopModule, AccidentModule, RouteModule, VehicleModule],
+    });
+
+  SwaggerModule.setup('/int/v1/docs/swagger', app, internalDocumentFactory, {
+    jsonDocumentUrl: '/int/v1/docs/openapi.json',
+  });
+
+  SwaggerModule.setup('/ext/v1/docs/swagger', app, externalDocumentFactory, {
+    jsonDocumentUrl: '/ext/v1/docs/openapi.json',
+  });
```
tchojnacki, line 51: **nit:** We do almost the same thing twice, can this be extracted to a function? e.g. `setupSwagger` with `title`, `description`, `version`, `modules` and `path` arguments?

implementation/leprechaun/src/shared/api/generic-paginated.dto.ts
```diff
@@ -0,0 +1,30 @@
+import { ApiProperty } from '@nestjs/swagger';
+
+export class PaginatedDto<TDto> {
```
tchojnacki, line 4: **question:** How do we determine whether the current page is the last? I genuinely can't recall what we decided on, and it's not written down in [in the contracts](https://github.com/PWR-ACS-SE-24/SoftwareSystemDesign/blob/main/documentation/contracts.md#pagination).

implementation/leprechaun/src/shared/api/generic-paginated.ts
```diff
@@ -0,0 +1,26 @@
+import { applyDecorators, Type } from '@nestjs/common';
+import { ApiExtraModels, ApiOkResponse, ApiQuery, getSchemaPath } from '@nestjs/swagger';
+import { PaginatedDto } from './generic-paginated.dto';
+
+export const ApiPaginatedResponse = <TModel extends Type>(model: TModel) => {
```
tchojnacki: **question:** since `TModel` is never used in the body, is there any benefit to using it instead of:
```suggestion
export const ApiPaginatedResponse = (model: Type) => {
```
?

implementation/leprechaun/src/shared/api/pipes.ts
```diff
@@ -0,0 +1,22 @@
+import { ParseUUIDPipe, ValidationPipe } from '@nestjs/common';
+import { SchemaMismatchException } from './http-exceptions';
+
+export const UUIDPipe = new ParseUUIDPipe({
+  version: '7',
+  exceptionFactory: (err) => new SchemaMismatchException(err as any),
```
tchojnacki: **issue:** `as any` is unnecessary, you are passing `err: string` to a `string` argument and cast the `string` to `any` to make sure it can be passed to a `string`?
```suggestion
  exceptionFactory: (err) => new SchemaMismatchException(err),
```

implementation/leprechaun/src/shared/api/pipes.ts
```diff
@@ -0,0 +1,22 @@
...
+  forbidNonWhitelisted: true,
+  exceptionFactory: (err) => new SchemaMismatchException(err as any),
+});
+
+export const ValidateUpdatePipe = new ValidationPipe({
+  transform: true,
+  whitelist: true,
+  forbidNonWhitelisted: true,
+  skipMissingProperties: true,
+  exceptionFactory: (err) => new SchemaMismatchException(err as any),
```
tchojnacki: **issue:** On the other hand, here `as any` is necessary to pass the typecheck, but it's because you shouldn't pass `err` here. `err` is of type `ValidationError[]`, yet you pretend that it is a `string` breaking type safety of all `SchemaMismatchException` consumers and possibly even leaking sensitive info? (I haven't checked the implementation of `HttpException` but we definitely don't want stacktraces etc. there.

You either have to find a way to compose an actual error string from `ValidationError[]` or if you are lazy, I'd rather have `new SchemaMismatchException("")` rather than `err as any`.

implementation/leprechaun/src/stop/controller/stop.controller.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  @Get('/')
+  @ApiPaginatedResponse(StopDto)
+  async getAllStops(
+    @Query('page') page_: number = 0,
+    @Query('size') size_: number = 50,
+    // @Query('filter') TODO: add filter
+  ): Promise<PaginatedDto<StopDto>> {
+    const { size, page } = PaginatedDto.sanitizePagination(size_, page_);
+    const { stops, total } = await this.stopService.listAll(size, page);
```
tchojnacki: **note:** Same as before.

implementation/leprechaun/src/stop/controller/stop.dto.ts
```diff
@@ -0,0 +1,37 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { Stop } from '../database/stop.entity';
+import { CreateStopDto } from './stop-create.dto';
+
+// TODO: add Lines to StopDto
+export class StopDto extends CreateStopDto {
```
tchojnacki: **nit:** Wouldn't it be better to have `StopDto` be the canonical definition and use [`OmitType`](https://docs.nestjs.com/openapi/mapped-types#omit) to define `CreateStopDto`?

implementation/leprechaun/src/test-setup.ts
```diff
@@ -0,0 +1,12 @@
+import { MetadataStorage, MikroORM } from '@mikro-orm/core';
+import { testConfig } from './config/mikro-orm.test.config';
+
+let _database: MikroORM;
+
+export default async (globalConfig, projectConfig) => {
+  const database = await MikroORM.init(testConfig);
+  await database.schema.createSchema();
+  globalThis.database = database;
```
tchojnacki, line 11: I hate how we're using global state for this, but I believe that you made sure this is necessary.

implementation/leprechaun/src/test-setup.ts
```diff
@@ -0,0 +1,12 @@
+import { MetadataStorage, MikroORM } from '@mikro-orm/core';
+import { testConfig } from './config/mikro-orm.test.config';
+
+let _database: MikroORM;
```
tchojnacki: **question:** Why? It's never used.

implementation/leprechaun/src/vehicle/controller/vehicle.controller.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  @Get('/')
+  @ApiPaginatedResponse(VehicleDto)
+  async getAllVehicles(
+    @Query('page') page_: number = 0,
+    @Query('size') size_: number = 50,
+    // @Query('filter') TODO: add filter
+  ): Promise<PaginatedDto<VehicleDto>> {
+    const { size, page } = PaginatedDto.sanitizePagination(size_, page_);
+    const { vehicles, total } = await this.vehicleService.listAll(size, page);
```
tchojnacki: **note:** Same as before.

implementation/leprechaun/src/vehicle/service/vehicle.service.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  async deleteVehicleById(vehicleId: string): Promise<void> {
+    // As per the requirements, we don't delete, but rather set isActive to false
+    const updated = await this.vehicleRepository.nativeUpdate({ id: vehicleId }, { isActive: false });
+    if (!updated) throw new NotFoundException(vehicleId);
+  }
+
+  async updateVehicleById(vehicleId: string, updateVehicle: UpdateVehicleDto): Promise<Vehicle> {
+    // TODO: switch all routes to use new vehicle when updating
```
tchojnacki: **question:** Why this TODO? Isn't the current behaviour correct? 🤔 If we unpin the previous vehicle from all of its associations, might as well hard-delete it.

implementation/leprechaun/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "baseUrl": "./",
+    "incremental": true,
+    "skipLibCheck": true,
+    "strict": true,
+    "strictNullChecks": false,
+    "noImplicitAny": false,
+    "strictBindCallApply": false,
+    "forceConsistentCasingInFileNames": false,
+    "noFallthroughCasesInSwitch": false
+  }
```
tchojnacki: Wait, so you enabled `strict`, which is a combination of multiple typechecking options and then manually disabled almost all of the included options, making `strict` useless? 😠

tchojnacki: I've enabled `strict`, removed your `: false` checks and ran type checking. And I can confirm that all of the errors you are getting are not false flags, and are caused by possibly unsafe operations.

implementation/leprechaun/src/line/controller/line-create.dto.ts
mlodybercik: In this very case *no*. `Create` recveives `string[]`, and `LineDto` has `StopDto` instead of `string[]`.

mlodybercik: Nvm. It can.

implementation/leprechaun/src/line/controller/line.controller.ts
mlodybercik: I did start implementing it on other branch before while waiting for a review, but I'll do in this one. Thanks for the sources!

implementation/leprechaun/src/shared/api/generic-paginated.dto.ts
mlodybercik: We discussed it but I don't remember if we settled on anything in particular.

implementation/leprechaun/src/test-setup.ts
mlodybercik: I need this to create schema for testing and tear it down afterwards. `Jest` fucks with something in the background, and I can't just recreate connection from scratch and drop schema. I did not debug much, but it is the only way that I know works.

implementation/leprechaun/src/vehicle/service/vehicle.service.ts
mlodybercik: Yeah, it was before consulting the propper behaviour with you.

implementation/leprechaun/tsconfig.json
mlodybercik: `nestjs` asks when creating project, whether to enable strict checking, and those values came from that. I didn't know tbh that strict is just alias for enabling some options. Seems kinda uneccessary for `nestjs` to set them explicitly instead of setting `strict` to false.

implementation/leprechaun/src/shared/api/pipes.ts
mlodybercik: I believe that these are remains of tests I did for debugging something with pipes.

implementation/leprechaun/src/shared/api/pipes.ts
mlodybercik: Same as above, I believe that these are remains of debugging. These were not even passed further to be included in any string or anything. <sub>remember, bug is just an undocumented feature!</sub>


`HttpException` exposes message as `string | Record<string, unknown>`. I'll use that to pass any neccessary non-critical data to exception handler to possibly enhance error description whenever it makes sense.

implementation/leprechaun/src/line/test/line.controller.spec.ts
```diff
@@ -0,0 +1,268 @@
...
+import { testConfig } from '../../config/mikro-orm.test.config';
+import { HttpExceptionFilter, InternalExceptionFilter } from '../../shared/api/http-exception.filter';
+import { SharedModule } from '../../shared/shared.module';
+import { Stop } from '../../stop/database/stop.entity';
+import { StopService } from '../../stop/service/stop.service';
+import { CreateLineDto } from '../controller/line-create.dto';
+import { LineController } from '../controller/line.controller';
+import { Line } from '../database/line.entity';
+import { StopLineMapping } from '../database/stop-line-mapping.entity';
+import { LineService } from '../service/line.service';
```
tchojnacki, line 15: **nit:** You might want to set up [absolute paths](https://medium.com/@bhusal_rj/configuring-absolute-path-in-nest-js-bc34f55425ef) to get rid of those `../../`

implementation/leprechaun/src/shared/api/http-exceptions.ts
```diff
@@ -0,0 +1,73 @@
...
+
+interface AppError {
+  kind: string;
+  code: number;
+  messageEn: string;
+  messagePl?: string;
+}
+
+export class HttpExceptionDto implements AppError {
+  @ApiProperty({ description: 'Error kind', enum: () => Object.values(ErrorKind).map((v) => v[0]), isArray: false })
```
tchojnacki: **question:** Is `isArray` `true` by default?

implementation/leprechaun/src/shared/api/http-exceptions.ts
```diff
@@ -0,0 +1,73 @@
...
+      messagePl = `Nie znaleziono zasobu o id {${message.details}}.`;
+    }
+    return new HttpExceptionDto(ErrorKind.NotFoundException, messageEn, messagePl);
+  },
+
+  InternalServerError: (_) =>
+    new HttpExceptionDto(ErrorKind.InternalServerError, 'Internal server error', 'Wewnętrzny błąd serwera'),
+};
+
+export function mapException<T extends HttpException>(error: T): AppError {
```
tchojnacki: **question:** What benefit does
```
export function mapException<T extends HttpException>(error: T): AppError {
```
give over
```
export function mapException(error: HttpException): AppError {
```

implementation/leprechaun/src/shared/api/pagination.decorator.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const Paginated = (limitOverwrite = DEFAULT_LIMIT) =>
+  createParamDecorator(
+    (_: unknown, context: ExecutionContext) => {
+      const request = context.switchToHttp().getRequest();
+
+      const page = parseInt(request.query.page, 10) || 0;
+      const size = parseInt(request.query.size, 10) || limitOverwrite;
+
+      return <Pagination>{
```
tchojnacki, line 20: **nit:** TypeScript is structurally typed, so this cast does absolutely nothing.

tchojnacki (approved):
**todo:** Set up CI/CD.

implementation/leprechaun/src/shared/api/http-exceptions.ts
mlodybercik: It is not, argument that I want to set it explicitly is stupid, I'll change that before merging.

https://docs.nestjs.com/openapi/types-and-parameters#enums

implementation/leprechaun/src/line/test/line.controller.spec.ts
mlodybercik: I'll do that in another smaller PR.

implementation/leprechaun/src/shared/api/pagination.decorator.ts
mlodybercik: I just want to make sure that TS shows me this:
image
instead of this
image

PR #49, opened 2024-12-29
PR: Account CRUD for Jobberknoll

tchojnacki:


implementation/jobberknoll/api/ext/routes/get-account-by-id-route.ts
```diff
@@ -0,0 +1,50 @@
...
+      SchemaMismatchDto,
+      "The request data did not align with the schema.",
+    ),
+  },
+});
+
+export function getAccountByIdHandler(
+  getAccountById: GetAccountByIdUseCase,
+): RouteHandler<typeof getAccountByIdRoute> {
+  return authorize("admin", async (c) => {
```
mlodybercik, line 42: **thought**: How hard would it be to make `authorize` accept multiple roles? I've checked with E2 and acording to the document you dont have to implement that.

implementation/jobberknoll/deno.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "0.2.0",
+  "version": "0.3.0",
```
mlodybercik, line 2: **thought**: It's not deployed yet, why do you bother?

mlodybercik (approved):
dr-house-house

implementation/jobberknoll/api/ext/routes/get-account-by-id-route.ts
tchojnacki:
```diff
  export function authorize<R extends RouteConfig>(
-   expectedRole: UserRole,
+   expectedRoles: UserRole[],
    handler: RouteHandler<R>,
  ): RouteHandler<R> {
    return (c, next) => {
      const userRole = c.req.header("jp-user-role");
-     if (expectedRole !== userRole) {
+     if (!expectedRoles.includes(userRole as UserRole)) {
```

Although I won't do it, since I wouldn't ever use it.

implementation/jobberknoll/deno.json
tchojnacki: No huge benefits honestly. It lets me see which version I'm running in both the console and both OpenAPI UIs, which isn't really important, but saved me once. And whenever this will actually communicate with other services (starting with Feather which I also have to write :c) it will get more useful.

PR #50, opened 2024-12-30
PR: Change account delete from soft to hard

tchojnacki:


implementation/jobberknoll/app/use-cases/create-account-use-case.ts
```diff
@@ -40,8 +40,7 @@ export class CreateAccountUseCase extends UseCase<
       fullName: req.fullName,
       email: req.email.toLowerCase(),
       hashedPassword: req.password, // TODO: hash password
-      isActive: true,
-      lastModified: Date.now(),
+      lastModified: Math.floor(Date.now() / 1000),
```
tchojnacki, line 43: JWT uses seconds instead of milliseconds, whoops.

implementation/jobberknoll/infra/account-repo/postgres-account-repo.ts
```diff
@@ -64,20 +24,22 @@ export class PostgresAccountRepo implements AccountRepo {
     const pool = new Pool(connectionString, POOL_SIZE, true);
 
     using client = await pool.connect();
+    await client.queryObject`
+      DO $$ BEGIN
+        CREATE TYPE ACCOUNT_TYPE AS ENUM ('admin', 'driver', 'inspector', 'passenger');
```
tchojnacki, line 29: There is no `CREATE TYPE IF NOT EXISTS` so we have to effectively do `try { create_type() } catch {}` :c

mlodybercik: Wont't creating a schema be handled by outside means anyways?

mlodybercik (approved):
The only thing turning from soft to hard is me.
thumbs-up-cat

implementation/jobberknoll/infra/account-repo/postgres-account-repo.ts
tchojnacki: Thanks for the idea, let's write it down and move on.

PR #52, opened 2025-01-04
PR: Jobberknoll logging, environment and infrastructure improvements

tchojnacki:


implementation/jobberknoll/api/ext/ext-controller.ts
```diff
@@ -1,20 +1,18 @@
...
 export class ExtController implements Controller {
   public constructor(private readonly service: Service, private readonly logger: Logger) {}
 
-  public get prefix(): string {
-    return "/ext/v1";
-  }
+  public readonly prefix = "/ext/v1";
 
   public get routes(): OpenAPIHono {
-    const app = new OpenAPIHono({ defaultHook })
```
tchojnacki, line 17: **note:** `createOpenAPIHono` sets up a `new OpenAPIHono` with the `defaultHook` as well as an error handler (thus `configureErrorHandler` below is removed) and two new shared middlewares.

implementation/jobberknoll/app/interfaces/logger.ts
```diff
@@ -0,0 +1,48 @@
+import type { UUID } from "@jobberknoll/core/shared";
```
tchojnacki, line 1: **note:** Mostly moved from `/app/shared/logger.ts`, Git doesn't recognize the move, since the implementation was also refactored to use `extends` instead of `implements` (but also recognizes an analogous change in `dev-logger.ts`, Git is weird).

implementation/jobberknoll/app/use-cases/use-case.ts
```diff
@@ -10,7 +10,7 @@ export abstract class UseCase<Req, Res, Err> {
 
     const res = await this.handle(req);
 
-    const tags = isOk(res) ? { res: res.value } : { err: res.value };
+    const tags = isOk(res) ? { ok: res.value } : { err: res.value };
```
tchojnacki, line 13: **note:** I somehow thought `Result` has variants of `Res` and `Err` instead of `Ok` and `Err`, whoops.

implementation/jobberknoll/README.md
mlodybercik, line 1: Great idea, we should do it all, across all of our microservices.

mlodybercik (approved):
Most/all of the changes are QOL or improvements in general. As always I had to read into the code to understand what it does, but as far I know, it looks good.

PR #53, opened 2025-01-06
PR: Improve and simplify instrumentation, add missing tests

tchojnacki:
1. **Problem:** The logging parameters `logger: Logger` and `requestId: UUID` were inconsequently passed around, sometimes they were the first argument, sometimes the last, sometimes in between. Two sensible answers would be to place them at the end or at the start.
    - **Solution:** Place all logging parameters at the start, this is more sensible, since a) variadic arguments can be only placed at the end so placing logger at the end would break it, b) when calling the function multiple time, placing common arguments at the start produces nicer aligment, c) that's what [opentracing](https://github.com/opentracing/opentracing-go#creating-a-span-given-an-existing-go-contextcontext) uses, d) all other services use the `Logger`, so passing it last would be weird. Thus **all `logger` and `requestId`** params should be first in function calls.
2. **Problem:** Request IDs passed around can be mistaken with other UUIDs and the request context can't be extended without major refactorings.
    - **Solution:** Replace all `requestId: UUID` arguments with `{ requestId }: Ctx` arguments.
3. **Problem:** Request IDs are not passed through to the `AccountRepo`, since `UseCase` trims request context. The code for instrumenting calls (in use cases and infra) is duplicated.
    - **Solution:** Create new `instrument` and `propagate` functions in `Logger`, then `propagate` use case handlers and `instrument` infra handlers.
4. **Problem:** Request context are not nicely extractable from HTTP requests.
    - **Solution:** Bind context to Hono's request to make it easy to access from all handlers.
5. **Problem:** Error DTO descriptions are duplicated all around, since DTO description and response descriptions are always the same.
    - **Solution:** Export response schemas for error DTOs to make the code more DRY. 
6. **Problem:** There is a ton of missing tests.
    - **Solution:** A bunch of new tests added, bringing the total to almost 150 tests with a ~90% coverage.

### DIFF
--- a/implementation/jobberknoll/app/security/hash.test.ts
+++ b/implementation/jobberknoll/app/security/hash.test.ts
@@ -1,0 +1,16 @@
+import { assertEquals } from "@std/assert";
+import { sha256 } from "./hash.ts";
+
+Deno.test("sha256 should return the correct hash", async () => {
+  for (
+    const [input, expected] of [
+      ["", "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"],
+      ["test", "9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08"],
+      ["hello world", "b94d27b9934d3e08a52e52d7da7dabfac484efe37a5380ee9088f7ace2efcde9"],
+    ]
+  ) {
+    const actual = await sha256(input);
+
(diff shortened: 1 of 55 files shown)


implementation/jobberknoll/app/security/hash.test.ts
mlodybercik, line 1: lol

mlodybercik (approved):

PR #54, opened 2025-01-08
PR: Add authguard and endpoints API to `leprechaun`

mlodybercik:


implementation/leprechaun/README.md
```diff
@@ -0,0 +1,67 @@
...
+| `GET    /ext/v1/accidents`     | 🟥         |
+| `GET    /ext/v1/accidents/:id` | 🟥         |
+| `POST   /ext/v1/accidents`     | 🟥         |
+| `PATCH  /ext/v1/accidents/:id` | 🟥         |
+
+## TODOs
+
+- Set up CI/CD
+- Most DTOs contain repeating fields like `id`, `isActive`. Create a base class that contains these fields and extend it in all DTOs.
+- We propably need explicit transactins in some places when doing more than single operation.
```
tchojnacki:
```suggestion
- We probably need explicit transactins in some places when doing more than single operation.
```

implementation/leprechaun/src/internal/controller/healthcheck.dto.ts
```diff
@@ -1,9 +1,10 @@
 import { ApiProperty } from '@nestjs/swagger';
 
-export type HealthcheckStatus = 'UP' | 'DOWN' | 'UNKNOWN' | 'OUT_OF_SERVICE';
+export const HealthcheckValues = ['UP', 'DOWN', 'UNKNOWN', 'OUT_OF_SERVICE'] as const;
```
tchojnacki: **nit:** Weird to have it cased this way.

implementation/leprechaun/src/internal/controller/routes.dto.ts
```diff
@@ -0,0 +1,33 @@
+import { RequestMethod } from '@nestjs/common';
+import { ApiProperty } from '@nestjs/swagger';
+import { Permissions, RoutePermissions } from '../service/auth.guard';
+
+export class RouteDto {
```
tchojnacki: **suggestion:** Wouldn't it be more convenient to name it `EndpointDto`? I purposefully named the path `/endpoints` instead of `/routes` even though the second one seemed more natural to me, because we already have a `Route` entity in the domain model and it could cause confusion.

implementation/leprechaun/src/internal/controller/routes.dto.ts
tchojnacki, line 1: **issue:** Please document it briefly in the `contracts.md` before we all do divergent solutions. There's even a section with a `TODO` ready there.

implementation/leprechaun/src/internal/service/auth.guard.ts
```diff
@@ -0,0 +1,77 @@
...
+  UseGuards,
+} from '@nestjs/common';
+import { METHOD_METADATA, PATH_METADATA } from '@nestjs/common/constants';
+import { Reflector } from '@nestjs/core';
+import { ApiForbiddenResponse, ApiHeader, ApiUnauthorizedResponse } from '@nestjs/swagger';
+
+const USER_ROLE_HEADER = 'jp-user-role';
+const ROLES_METADATA = 'roles';
+
+// Could do this as binary enum flags, but this is more readable
```
tchojnacki, line 19: 👍 no need to prematurely optimize

implementation/leprechaun/src/internal/service/auth.guard.ts
```diff
@@ -0,0 +1,77 @@
...
+  return applyDecorators(
+    SetMetadata(ROLES_METADATA, roles),
+    UseGuards(AuthGuard),
+    ApiHeader({
+      name: USER_ROLE_HEADER,
+      description: 'User role header',
+      required: true,
+      // show required roles instead of all
+      enum: roles,
+    }),
```
tchojnacki:
```suggestion
export const RequiredPermissions = (role: RoutePermissions, ...roles: RoutePermissions[]) => {
  return applyDecorators(
    SetMetadata(ROLES_METADATA, roles),
    UseGuards(AuthGuard),
    ApiHeader({
      name: USER_ROLE_HEADER,
      description: 'User role header',
      required: true,
      // show required roles instead of all
      enum: [role, ...roles],
    }),
```

**suggestion:** I saw you trying to enforce, that at the `roles` array has at least one element, emitting warnings otherwise. You can check it at compile time instead of runtime, if you change the signature to something like this.

implementation/leprechaun/src/internal/service/auth.guard.ts
```diff
@@ -0,0 +1,77 @@
...
+import { METHOD_METADATA, PATH_METADATA } from '@nestjs/common/constants';
+import { Reflector } from '@nestjs/core';
+import { ApiForbiddenResponse, ApiHeader, ApiUnauthorizedResponse } from '@nestjs/swagger';
+
+const USER_ROLE_HEADER = 'jp-user-role';
+const ROLES_METADATA = 'roles';
+
+// Could do this as binary enum flags, but this is more readable
+export const Permissions = ['guest', 'passenger', 'driver', 'admin', 'inspector'] as const;
+export type RoutePermissions = (typeof Permissions)[number];
```
tchojnacki:
```suggestion
export type RoutePermission = (typeof Permissions)[number];
```

**nit:** An element in an array of permissions, would be a "permission", not "permissions". It is pretty confusing to see `"admin"` have the type of `Permissions` instead of `Permission`.

implementation/leprechaun/src/internal/service/auth.guard.ts
```diff
@@ -0,0 +1,77 @@
...
+      const httpMethod = this.reflector.get<number>(METHOD_METADATA, methodHandler);
+      const controllerPath = this.reflector.get<string>(PATH_METADATA, context.getClass());
+      const methodPath = this.reflector.get<string>(PATH_METADATA, methodHandler);
+      this.logger.warn(`No roles defined for ${RequestMethod[httpMethod]} ${controllerPath + methodPath}`);
+      throw new ForbiddenException();
+    }
+
+    // If the role is not in the list of roles, throw a ForbiddenException
+    if (!roles.includes(role)) {
+      throw new ForbiddenException();
```
tchojnacki:
```suggestion
      return false;
    }

    // If the role is not in the list of roles, throw a ForbiddenException
    if (!roles.includes(role)) {
      return false;
```

**question:** Wouldn't this be equivalent?

tchojnacki (approved):
LGTM, left some suggestions 👍

implementation/leprechaun/src/internal/controller/routes.dto.ts
mlodybercik: I would probably notice that when implementing `Route`.

implementation/leprechaun/src/internal/service/auth.guard.ts
mlodybercik: Neat

implementation/leprechaun/src/internal/service/auth.guard.ts
mlodybercik: For forbidden yes. For unauthorized no. I saw somewhere examples dropping plain `ForbiddenExceptions` here so I copied it.

Gf6BWruXwAA5sdL

PR #55, opened 2025-01-09
PR: `leprechaun` add CI/CD, add `AbstractIdDTO`, add `@app`

mlodybercik:
Did some TODO's. Namely:
- make imports absolute by adding `@app` to path,
- add CI/CD,
- move common `isActive` and `id` to common abstract DTO.


implementation/leprechaun/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+FROM base AS builder
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+WORKDIR /app
+COPY . /app
+RUN corepack enable && pnpm fetch --prod && pnpm run build
+
+FROM base AS runner
+WORKDIR /app
+RUN apk add --no-cache curl
```
jakubzehner: You can move this to line number 2, then it will cached and won't be installed again and again with every change in the code.

implementation/jobberknoll/compose.yml
```diff
@@ -8,7 +8,8 @@ services:
       SERVER_PORT: "8000"
       DATABASE_URL: "postgres://postgres:password@jobberknoll-db:5432/postgres"
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:8000/int/v1/health"]
+      test:
+        ["CMD-SHELL", "curl -f http://localhost:$$SERVER_PORT/int/v1/health"]
```
tchojnacki, line 12: ❤️

implementation/leprechaun/README.md
```diff
@@ -61,7 +61,5 @@
 
 ## TODOs
 
-- Set up CI/CD
-- Most DTOs contain repeating fields like `id`, `isActive`. Create a base class that contains these fields and extend it in all DTOs.
-- We probably need explicit transactins in some places when doing more than single operation.
+- Ultimately, we will HAVE to implement some kind AWS secret-manager support.
```
tchojnacki:
```suggestion
- Ultimately, we will HAVE to implement some kind of secret-manager support.
```
**nit:**

.github/workflows/leprechaun.yml
```diff
@@ -0,0 +1,61 @@
...
+        with:
+          node-version: "22.x"
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v4.0.0
+        with:
+          version: "10"
+
+      - name: Install and test
+        run: pnpm it
```
tchojnacki, line 61: **question:** Where does `pnpm it` come from? I couldn't find in in PNPM documentation and neither in the `package.json`.

implementation/leprechaun/src/config/config.ts
```diff
@@ -0,0 +1,15 @@
+interface AppConfig {
+  PROD: boolean;
+  LEPRECHAUN_APP_PORT: number;
+  LEPRECHAUN_DATABASE_HOST: string;
+  LEPRECHAUN_DATABASE_DBNAME: string;
+}
+
+const config: AppConfig = {
+  PROD: process.env.NODE_ENV === 'production',
+  LEPRECHAUN_APP_PORT: parseInt(process.env.LEPRECHAUN_APP_PORT!) ?? 3000,
```
tchojnacki: **nit:** `parseInt` always returns a `number`, and `??` never does anything to a number. If you wanted to use `3000` if the result of `parseInt` is `NaN` this unfortunately won't work. You can use `||` which I'd normally discourage here instead.

implementation/leprechaun/src/config/config.ts
```diff
@@ -0,0 +1,15 @@
...
+  LEPRECHAUN_DATABASE_HOST: process.env.LEPRECHAUN_DATABASE_HOST! ?? 'postgresql://postgres:test@localhost:5432',
+  LEPRECHAUN_DATABASE_DBNAME: process.env.LEPRECHAUN_DATABASE_HOST! ?? 'leprechaun',
```
tchojnacki: On the other hand `??` is perfect here, but I don't know why you'd need `!` before them.

tchojnacki (approved):

.github/workflows/leprechaun.yml
mlodybercik: `it` is an alias for `install-test`. Sparse documentation [here](https://pnpm.io/cli/install-test), and in the CLI:
```
Version 9.12.2 (compiled to binary; bundled Node.js v22.9.0)
Usage: pnpm [command] [flags]
       pnpm [ -h | --help | -v | --version ]

Manage your dependencies:
      add                  Installs a package and any packages that it depends
                           on. By default, any new package is installed as a
                           prod dependency
      import               Generates a pnpm-lock.yaml from an npm
                           package-lock.json (or npm-shrinkwrap.json) file
   i, install              Install all dependencies for a project
  it, install-test         Runs a pnpm install followed immediately by a pnpm
                           test
...
```

implementation/leprechaun/src/config/config.ts
mlodybercik: Without the `!` in `parseInt` TS would complain about undefined not being assignable to `string`, and also, before adding `??` i just had `<name>: <env>` and it would also complain about undefined not being assignable to whatever type `AppConfig` needed.

PR #58, opened 2025-01-10
PR: Infrastructure health for Jobberknoll

tchojnacki:


implementation/jobberknoll/app/interfaces/logger.ts
```diff
@@ -29,7 +30,17 @@ const LEVELS = {
   error: 50,
 };
 
-export abstract class Logger {
+export abstract class Logger implements Monitorable {
+  public health(): Promise<ComponentHealth> {
```
tchojnacki, line 34: **note:** Adding health to loggers might seem weird, but it's very common to have a logger send data to a remote aggregator or save to file, or anything like that, which would actually make it able to be `DOWN`.

mlodybercik (approved):
mfw
subway-surfers-icegif-5

PR #60, opened 2025-01-11
PR: `leprechaun` add `Route` functionality

mlodybercik:
- adds another CRUD, 
- adds custom validator for dates,
- fixes some inconsistencies in controllers not showing propper responses

### DIFF
--- a/implementation/leprechaun/src/route/controller/route-create.dto.ts
+++ b/implementation/leprechaun/src/route/controller/route-create.dto.ts
@@ -1,16 +1,16 @@
 import { IsLaterThan } from '@app/shared/api/date-larger.validator';
 import { ApiProperty, PartialType, PickType } from '@nestjs/swagger';
-import { IsNotEmpty, IsOptional, IsRFC3339, IsUUID } from 'class-validator';
+import { IsISO8601, IsNotEmpty, IsOptional, IsUUID } from 'class-validator';
 import { RouteDto } from './route.dto';
 
 export class CreateRouteDto extends PickType(RouteDto, ['startTime', 'endTime']) {
-  @IsRFC3339()
+  @IsISO8601()
   @IsNotEmpty()
   readonly startTime: string;
 
-  @IsRFC3339()
+  @IsISO8601()
   @IsNotEmpty()
-  @IsLaterThan('startTime')
+  @IsLaterThan<CreateRouteDto>('startTime')
   readonly endTime: string;
 
   @IsNotEmpty()
@@ -45,11 +45,11 @@ export class CreateRouteDto extends PickType(RouteDto, ['startTime', 'endTime'])
 }
 
 export class UpdateRouteDto extends PartialType(PickType(CreateRouteDto, ['line', 'vehicle'])) {
-  @IsRFC3339()
+  @IsISO8601()
   @IsOptional()
   readonly startTime?: string;
 
-  @IsRFC3339()
+  @IsISO8601()
   @IsOptional()
   readonly endTime?: string;
 
--- a/implementation/leprechaun/src/route/controller/route-create.dto.ts
+++ b/implementation/leprechaun/src/route/controller/route-create.dto.ts
@@ -1,0 +1,61 @@
+import { IsLaterThan } from '@app/shared/api/date-larger.validator';
+import { ApiProperty, PartialType, PickType } from '@nestjs/swagger';
+import { IsNotEmpty, IsRFC3339, IsUUID } from 'class-validator';
+import { RouteDto } from './route.dto';
+
+export class CreateRouteDto extends PickType(RouteDto, ['startTime', 'endTime']) {
+  @IsRFC3339()
+  @IsNotEmpty()
+  readonly startTime: string;
+
+  @IsRFC3339()
+  @IsNotEmpty()
+  @IsLaterThan('startTime')
+  readonly endTime: string;
+
+  @IsNotEmpty()
+  @IsUUID('7')
+  @ApiProperty({
+    description: 'Route line',
+    type: 'string',
+    format: 'uuid',
+    nullable: false,
+    example: 'a7f8b7c0-2b4f-7b6e-9e1e-6f2b1b4c4f4d',
+  })
+  readonly line: string;
+
+  @IsNotEmpty()
+  @IsUUID('7')
+  @ApiProperty({
+    description: 'Vehicle on route',
+    type: 'string',
+    format: 'uuid',
+    nullable: false,
+    example: '0194137a-3faf-70fb-8388-6586b32d9c8a',
+  })
+  readonly vehicle: string;
+
+  constructor(startTime: string, endTime: string, line: string, vehicle: string) {
+    super();
+    this.startTime = startTime;
+    this.endTime = endTime;
+    this.line = line;
+    this.vehicle = vehicle;
+  }
+}
+
+export class UpdateRouteDto extends PartialType(PickType(CreateRouteDto, ['line', 'vehicle'])) {
+  @IsRFC3339()
+  @IsNotEmpty()
+  readonly startTime: string;
+
+  @IsRFC3339()
+  @IsNotEmpty()
+  readonly endTime: string;
+
+  constructor(startTime: string, endTime: string, line: string, vehicle: string) {
+    super(line, vehicle);
+    this.startTime = startTime;
+    this.endTime = endTime;
+  }
+}
--- a/implementation/leprechaun/src/route/controller/route-create.dto.ts
+++ b/implementation/leprechaun/src/route/controller/route-create.dto.ts
@@ -1,6 +1,6 @@
 import { IsLaterThan } from '@app/shared/api/date-larger.validator';
 import { ApiProperty, PartialType, PickType } from '@nestjs/swagger';
-import { IsNotEmpty, IsRFC3339, IsUUID } from 'class-validator';
+import { IsNotEmpty, IsOptional, IsRFC3339, IsUUID } from 'class-validator';
 import { RouteDto } from './route.dto';
 
 export class CreateRouteDto extends PickType(RouteDto, ['startTime', 'endTime']) {
@@ -46,14 +46,14 @@ export class CreateRouteDto extends PickType(RouteDto, ['startTime', 'endTime'])
 
 export class UpdateRouteDto extends PartialType(PickType(CreateRouteDto, ['line', 'vehicle'])) {
   @IsRFC3339()
-  @IsNotEmpty()
-  readonly startTime: string;
+  @IsOptional()
+  readonly startTime?: string;
 
   @IsRFC3339()
-  @IsNotEmpty()
-  readonly endTime: string;
+  @IsOptional()
+  readonly endTime?: string;
 
-  constructor(startTime: string, endTime: string, line: string, vehicle: string) {
+  constructor(line: string, vehicle: string, startTime?: string, endTime?: string) {
     super(line, vehicle);
     this.startTime = startTime;
     this.endTime = endTime;
--- a/implementation/leprechaun/src/route/service/route.service.ts
+++ b/implementation/leprechaun/src/route/service/route.service.ts
@@ -1,9 +1,10 @@
 import { LineService } from '@app/line/service/line.service';
 import { Pagination } from '@app/shared/api/pagination.decorator';
 import { VehicleService } from '@app/vehicle/service/vehicle.service';
-import { EntityManager, EntityRepository } from '@mikro-orm/core';
+import { EntityRepository } from '@mikro-orm/core';
 import { InjectRepository } from '@mikro-orm/nestjs';
-import { BadRequestException, Injectable, NotFoundException } from '@nestjs/common';
+import { EntityManager } from '@mikro-orm/postgresql';
+import { BadRequestException, Injectable, InternalServerErrorException, NotFoundException } from '@nestjs/common';
 import { CreateRouteDto, UpdateRouteDto } from '../controller/route-create.dto';
 import { Route } from '../database/route.entity';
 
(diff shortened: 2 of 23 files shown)


implementation/leprechaun/src/route/controller/route-create.dto.ts
```diff
@@ -0,0 +1,61 @@
+import { IsLaterThan } from '@app/shared/api/date-larger.validator';
+import { ApiProperty, PartialType, PickType } from '@nestjs/swagger';
+import { IsNotEmpty, IsRFC3339, IsUUID } from 'class-validator';
+import { RouteDto } from './route.dto';
+
+export class CreateRouteDto extends PickType(RouteDto, ['startTime', 'endTime']) {
+  @IsRFC3339()
```
tchojnacki: **question:** Are we sure that this is the correct decorator after all? Since `@IsDateString()` is an alias for `@IsISO8601()` and not `@IsRFC3339()`, and [Date.prototype.toJSON](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-date.prototype.tojson) (which we need) returns [Date.prototype.toISOString](https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-date.prototype.tojson).

implementation/leprechaun/src/route/service/route.service.ts
```diff
@@ -0,0 +1,146 @@
...
+    });
+    const total = await this.routeRepository.count();
+
+    return { routes, total };
+  }
+
+  private async findRoutesInTimeRangeForVehicle(
+    vehicleId: string,
+    newStartTime: Date,
+    newOldTime: Date,
```
tchojnacki: **issue:** What? Shouldn't that be `newEndTime` instead of `newOldTime`?

implementation/leprechaun/src/route/service/route.service.ts
```diff
@@ -0,0 +1,146 @@
...
+    // prettier-ignore
+    return await this.routeRepository.find({
+      vehicle: vehicleId,
+      $or: [
+        /* 1) */ { startTime: { $gte: newStartTime, $lte: newOldTime }, endTime: {$gte: newOldTime} },
+        /* 2) */ { startTime: { $gte: newStartTime }, endTime: { $lte: newOldTime } },
+        /* 3) */ { startTime: { $lte: newStartTime }, endTime: { $gte: newStartTime, $lte: newOldTime } },
+      ],
+      isActive: true,
+    });
```
tchojnacki: Assuming per my last comment that you meant `newEndTime`.

**suggestion:** As long as both ranges are valid (`startTime` < `endTime`, should be enforced in the DB anyways; `newStartTime < newOldTime`, should be enforced by the validator), this can be massively simplified https://stackoverflow.com/a/325964.

implementation/leprechaun/src/route/service/route.service.ts
```diff
@@ -0,0 +1,146 @@
...
+    const vehicle = await this.vehicleService.getVehicleById(route.vehicle);
+
+    const startTime = new Date(route.startTime);
+    const endTime = new Date(route.endTime);
+
+    // this is very much redundant here because of validator, but good sanity check
+    if (new Date() > startTime || startTime > endTime) {
+      // TODO: move me to a common function for use with updateRoute
+      throw new BadRequestException({ details: 'Start time invalid' });
+    }
```
tchojnacki: **nit:** If this is already enforced by the validator, if these conditions are false here, it's the fault of the service and not the user, and thus a `5xx` error, not a `4xx`.

implementation/leprechaun/src/route/service/route.service.ts
```diff
@@ -0,0 +1,146 @@
...
+    // find all routes for the vehicle in the time range, if there are any, throw an error
+    if ((await this.findRoutesInTimeRangeForVehicle(vehicle.id, startTime, endTime)).length > 0) {
+      throw new BadRequestException({ details: 'Vehicle is already being used in this time range' });
+    }
+
+    // all good, create the route
+    const newRoute = new Route(startTime, endTime, line, vehicle);
+    await this.em.persistAndFlush(newRoute);
+    this.em.populate(newRoute, ['line.mappings.stop', 'vehicle']);
+    return newRoute;
```
tchojnacki: **issue:** Will this hold under load tests? You are asynchronously validating stuff in the service, which is in no way enforced in the DB and you aren't locking the DB in the meantime. What if two users (`VU1` and `VU2`) submit a route creation at the same time?

```
1. VU1 attempts to create Route1.
2. VU2 attempts to create Route2 (in the same time span as Route1).
3. System validates VU1's request and checks findRoutesInTimeRangeForVehicle, which is valid since the vehicle is not busy at that time.
4. System validates VU2's request and also does the check, which once again passes since at that point Route1 is still not saved to the DB.
5. System saves Route1 to the DB.
6. System saves Route2 to the DB.
7. We end up with invalid state in the DB.
```

Doing such separated check is nice if you want to present a cool error message to the user, but isn't sufficient unless you a) make sure the DB can't change underneath you through transactions or locking or something OR b) make sure what you're enforcing the constraint also in the DB.

For instance, I check emails this way -> there's a check for email availability and later I insert the user separately, but the email uniqueness is still enforced by the DB, so I get:
- if the user is nice -> nice error message (same in your case)
- if the user is naughty -> a 5XX error, with no integrity breach (different in your case, integrity gets broken)

implementation/leprechaun/src/route/service/route.service.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  async updateRoute(routeId: string, newRoute: UpdateRouteDto): Promise<Route> {
```
tchojnacki, line 92: **note:** Same issues as in `create`.

implementation/leprechaun/src/route/service/route.service.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    // find all routes for the vehicle in the time range
+    const plannedRoutes = await this.findRoutesInTimeRangeForVehicle(vehicle.id, startTime, endTime);
+
+    // if there are more than one route in the time range check whether its the same we are updating from
+    plannedRoutes
+      .filter((route) => route.id !== oldRoute.id)
+      .map((_) => {
+        throw new BadRequestException({ details: 'Vehicle is already being used in this time range' });
+      });
```
tchojnacki: **nit:** `map` shouldn't be used for side effects
```suggestion
      .forEach(() => {
        throw new BadRequestException({ details: 'Vehicle is already being used in this time range' });
      });
```

implementation/leprechaun/src/route/service/route.service.ts
```diff
@@ -0,0 +1,146 @@
...
+
```
tchojnacki: **suggestion:** Or better yet, do something like this.
```suggestion
    if (plannedRoutes.some((route) => route.id !== oldRoute.id)) {
      throw new BadRequestException({ details: 'Vehicle is already being used in this time range' });
    }

```

implementation/leprechaun/src/route/test/route.controller.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+import { EntityManager, MikroORM } from '@mikro-orm/core';
+import { MikroOrmModule } from '@mikro-orm/nestjs';
+import { BadRequestException } from '@nestjs/common';
+import { Test, TestingModule } from '@nestjs/testing';
+import { RouteController } from '../controller/route.controller';
+import { Route } from '../database/route.entity';
+import { RouteService } from '../service/route.service';
+
+function createTime(hours: number): Date {
+  return new Date(new Date().valueOf() + Math.floor(hours * 3600 * 1000));
```
tchojnacki: **question:** Why the `Math.floor`? Does it ever do something here?

implementation/leprechaun/src/route/test/route.controller.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+import { VehicleModule } from '@app/vehicle/vehicle.module';
+import { EntityManager, MikroORM } from '@mikro-orm/core';
+import { MikroOrmModule } from '@mikro-orm/nestjs';
+import { BadRequestException } from '@nestjs/common';
+import { Test, TestingModule } from '@nestjs/testing';
+import { RouteController } from '../controller/route.controller';
+import { Route } from '../database/route.entity';
+import { RouteService } from '../service/route.service';
+
+function createTime(hours: number): Date {
```
tchojnacki: **very nit:** The function name and params tells me nothing about what the function does without looking at the implementation. Maybe something like `dateNHoursFromNow` or `addHoursToCurrentTime` or `hoursFromNow`?

implementation/leprechaun/src/route/test/route.controller.spec.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    await expect(
+      controller.createRoute({
+        line: otherLine.id,
+        vehicle: vehicle.id,
+        startTime: createTime(0.5).toISOString(),
+        endTime: createTime(1.5).toISOString(),
+      }),
+    ).rejects.toThrow(BadRequestException);
+  });
```
tchojnacki: **note:** What I've talked about before, here you politely wait for the previous route to be created before attempting a new one, what if you do `controller.createRoute` N times through `Promise.all` instead?

implementation/leprechaun/src/shared/api/date-larger.validator.ts
```diff
@@ -0,0 +1,24 @@
...
+    registerDecorator({
+      name: 'isLaterThan',
+      target: object.constructor,
+      propertyName: propertyName,
+      constraints: [property],
+      options: validationOptions,
+      validator: {
+        validate(currentValue: any, args: ValidationArguments) {
+          const [relatedPropertyName] = args.constraints;
+          const relatedValue = (args.object as any)[relatedPropertyName];
```
tchojnacki, line 14: **nit:** If you pass an invalid `relatedPropertyName`, this will throw instead of returning `false`, not sure if it's intended.

implementation/leprechaun/src/route/test/route.controller.spec.ts
mlodybercik: I didn't care much to check how `Date` handles floating point numbers, so to be sure I did `floor`. It prolly does not change anything.

implementation/leprechaun/src/route/controller/route-create.dto.ts
mlodybercik: Yeah, I'll replace all occureances of `RFC3339` to `ISO8601`. These are simmilar enough so one can be parsed by the other most of the time (\*ba dum tss\*).

See [here](https://ijmacd.github.io/rfc3339-iso8601/).

implementation/leprechaun/src/route/service/route.service.ts
mlodybercik: Yeah, typo.

implementation/leprechaun/src/route/service/route.service.ts
mlodybercik: Looks like it can be optimized further by using `postgres`'s `OVERLAPS` operator.

Note 1: Mikro-ORM does not support `OVERLAPS/&&` for `tstzrange` required here.
Note 2: It's slower to use `&&` with two `tstzranges` than two comparisons (lol).

<details>
<code>
leprechaun=# explain
select
    &quot;r0&quot;.*,
    &quot;l1&quot;.&quot;id&quot; as &quot;l1__id&quot;,
    &quot;v2&quot;.&quot;id&quot; as &quot;v2__id&quot;
from
    &quot;route&quot; as &quot;r0&quot;
    inner join &quot;line&quot; as &quot;l1&quot; on &quot;r0&quot;.&quot;line_id&quot; = &quot;l1&quot;.&quot;id&quot;
    and &quot;l1&quot;.&quot;is_active&quot; = true
    inner join &quot;vehicle&quot; as &quot;v2&quot; on &quot;r0&quot;.&quot;vehicle_id&quot; = &quot;v2&quot;.&quot;id&quot;
    and &quot;v2&quot;.&quot;is_active&quot; = true
where
    &quot;r0&quot;.&quot;is_active&quot; = true
    and &quot;r0&quot;.&quot;vehicle_id&quot; = &#39;01945616-d5b8-7566-b423-848c29467425&#39;
    and &quot;r0&quot;.&quot;start_time&quot; &lt;= CURRENT_TIMESTAMP
    and &quot;r0&quot;.&quot;end_time&quot; &gt;= CURRENT_TIMESTAMP
    and &quot;r0&quot;.&quot;is_active&quot; = true;
 Nested Loop  (cost=0.30..45.95 rows=1 width=97)
   -&gt;  Nested Loop  (cost=0.15..37.77 rows=1 width=81)
         -&gt;  Seq Scan on route r0  (cost=0.00..29.57 rows=1 width=65)
               Filter: (is_active AND is_active AND (vehicle_id = &#39;01945616-d5b8-7566-b423-848c29467425&#39;::uuid) AND (start_time &lt;= CURRENT_TIMESTAMP) AND (end_time &gt;= CURRENT_TIMESTAMP))
         -&gt;  Index Scan using line_pkey on line l1  (cost=0.15..8.17 rows=1 width=16)
               Index Cond: (id = r0.line_id)
               Filter: is_active
   -&gt;  Index Scan using vehicle_pkey on vehicle v2  (cost=0.15..8.17 rows=1 width=16)
         Index Cond: (id = &#39;01945616-d5b8-7566-b423-848c29467425&#39;::uuid)
         Filter: is_active
(10 rows)
leprechaun=# explain
select
    &quot;r0&quot;.*,
    &quot;l1&quot;.&quot;id&quot; as &quot;l1__id&quot;,
    &quot;v2&quot;.&quot;id&quot; as &quot;v2__id&quot;
from
    &quot;route&quot; as &quot;r0&quot;
    inner join &quot;line&quot; as &quot;l1&quot; on &quot;r0&quot;.&quot;line_id&quot; = &quot;l1&quot;.&quot;id&quot;
    and &quot;l1&quot;.&quot;is_active&quot; = true
    inner join &quot;vehicle&quot; as &quot;v2&quot; on &quot;r0&quot;.&quot;vehicle_id&quot; = &quot;v2&quot;.&quot;id&quot;
    and &quot;v2&quot;.&quot;is_active&quot; = true
where
    &quot;r0&quot;.&quot;is_active&quot; = true
    and &quot;r0&quot;.&quot;vehicle_id&quot; = &#39;01945616-d5b8-7566-b423-848c29467425&#39;
    and tstzrange(&quot;r0&quot;.&quot;start_time&quot;, &quot;r0&quot;.&quot;end_time&quot;, &#39;[]&#39;) &amp;&amp; tstzrange(CURRENT_TIMESTAMP, CURRENT_TIMESTAMP, &#39;[]&#39;);

Nested Loop  (cost=0.30..48.12 rows=1 width=97)
   -&gt;  Nested Loop  (cost=0.15..39.95 rows=1 width=81)
         -&gt;  Seq Scan on route r0  (cost=0.00..31.75 rows=1 width=65)
               Filter: (is_active AND (vehicle_id = &#39;01945616-d5b8-7566-b423-848c29467425&#39;::uuid) AND (tstzrange(start_time, end_time, &#39;[]&#39;::text) &amp;&amp; tstzrange(CURRENT_TIMESTAMP, CURRENT_TIMESTAMP, &#39;[]&#39;::text)))
         -&gt;  Index Scan using line_pkey on line l1  (cost=0.15..8.17 rows=1 width=16)
               Index Cond: (id = r0.line_id)
               Filter: is_active
   -&gt;  Index Scan using vehicle_pkey on vehicle v2  (cost=0.15..8.17 rows=1 width=16)
         Index Cond: (id = &#39;01945616-d5b8-7566-b423-848c29467425&#39;::uuid)
         Filter: is_active
(10 rows)
</code>
</details>

implementation/leprechaun/src/route/service/route.service.ts
mlodybercik: Yeah, valid point. Race-condition is very much possible here. It can be easily omited by extending transactions (in `update` extending, in `create` adding one).

implementation/leprechaun/src/route/test/route.controller.spec.ts
mlodybercik: I am scared of doing `Promise.all` in tests, because it introduces *some* uncertainty in execution, and might indeed show some race-conditions 🙈.

implementation/leprechaun/src/shared/api/date-larger.validator.ts
mlodybercik: I made it check at compile-time by using types!

implementation/leprechaun/src/route/test/route.controller.spec.ts
tchojnacki: Yeah, I wouldn't actually add it here, just a hypothetical scenario.

It would be a great example of a [k6 functional test](https://grafana.com/docs/k6/latest/examples/functional-testing/) sometime in the future though.

tchojnacki (approved):
LGTM, nice job 👌

PR #61, opened 2025-01-11
PR: Implement "Self" use cases

tchojnacki:
obraz

### DIFF
--- a/implementation/jobberknoll/app/shared/schemas.ts
+++ b/implementation/jobberknoll/app/shared/schemas.ts
@@ -5,7 +5,7 @@ export const FullNameSchema: z.ZodString = z.string().min(1).max(255);
 export const EmailSchema: z.ZodString = z.string().email().max(255);
 
 // TODO: what are our password requirements?
-export const PasswordSchema: z.ZodString = z.string().min(8).max(64);
+export const PasswordSchema: z.ZodString = z.string().min(8).max(50);
 
 // TODO: what are our phone number constraints?
 export const PhoneNumberSchema: z.ZodNullable<z.ZodString> = z.string().regex(/^[0-9+ ]{1,16}$/).nullable();
--- a/implementation/jobberknoll/app/shared/schemas.ts
+++ b/implementation/jobberknoll/app/shared/schemas.ts
@@ -5,7 +5,7 @@ export const FullNameSchema: z.ZodString = z.string().min(1).max(255);
 export const EmailSchema: z.ZodString = z.string().email().max(255);
 
 // TODO: what are our password requirements?
-export const PasswordSchema: z.ZodString = z.string().min(8).max(50);
+export const PasswordSchema: z.ZodString = z.string().min(8).max(100);
 
 // TODO: what are our phone number constraints?
 export const PhoneNumberSchema: z.ZodNullable<z.ZodString> = z.string().regex(/^[0-9+ ]{1,16}$/).nullable();
--- a/implementation/jobberknoll/app/shared/schemas.ts
+++ b/implementation/jobberknoll/app/shared/schemas.ts
@@ -1,11 +1,11 @@
 import { z } from "@hono/zod-openapi";
 
-export const FullNameSchema = z.string().min(1).max(255);
+export const FullNameSchema: z.ZodString = z.string().min(1).max(255);
 
-export const EmailSchema = z.string().email().max(255);
+export const EmailSchema: z.ZodString = z.string().email().max(255);
 
 // TODO: what are our password requirements?
-export const PasswordSchema = z.string().min(8).max(64);
+export const PasswordSchema: z.ZodString = z.string().min(8).max(64);
 
 // TODO: what are our phone number constraints?
-export const PhoneNumberSchema = z.string().regex(/^[0-9+ ]{1,16}$/).nullable();
+export const PhoneNumberSchema: z.ZodNullable<z.ZodString> = z.string().regex(/^[0-9+ ]{1,16}$/).nullable();
--- a/implementation/jobberknoll/app/shared/schemas.ts
+++ b/implementation/jobberknoll/app/shared/schemas.ts
@@ -4,8 +4,8 @@ export const FullNameSchema: z.ZodString = z.string().min(1).max(255);
 
 export const EmailSchema: z.ZodString = z.string().email().max(255);
 
-// TODO: what are our password requirements?
-export const PasswordSchema: z.ZodString = z.string().min(8).max(100);
+// NOTE: at least one uppercase, at least one lowercase, at least 8 characters, at most 100 characters
+export const PasswordSchema: z.ZodString = z.string().regex(/^(?=.*[A-Z])(?=.*[a-z]).{8,100}$/);
 
-// TODO: what are our phone number constraints?
-export const PhoneNumberSchema: z.ZodNullable<z.ZodString> = z.string().regex(/^[0-9+ ]{1,16}$/).nullable();
+// NOTE: optional country code preceded by '+', then exactly 9 digits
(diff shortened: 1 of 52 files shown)


implementation/jobberknoll/app/use-cases/edit-account-password-use-case.test.ts
```diff
@@ -0,0 +1,142 @@
+import { accountMock, isErr, isOk } from "@jobberknoll/core/shared";
```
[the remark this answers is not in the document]
mlodybercik: 🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤🤤
21st-century-brain-rot

implementation/jobberknoll/app/use-cases/edit-account-phone-use-case.test.ts
```diff
@@ -0,0 +1,70 @@
+import type { Passenger } from "@jobberknoll/core/domain";
```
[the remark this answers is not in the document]
mlodybercik: homer-simpsons

implementation/jobberknoll/app/shared/schemas.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const FullNameSchema: z.ZodString = z.string().min(1).max(255);
+
+export const EmailSchema: z.ZodString = z.string().email().max(255);
+
+// TODO: what are our password requirements?
+export const PasswordSchema: z.ZodString = z.string().min(8).max(100);
+
+// TODO: what are our phone number constraints?
+export const PhoneNumberSchema: z.ZodNullable<z.ZodString> = z.string().regex(/^[0-9+ ]{1,16}$/).nullable();
```
mlodybercik: **thought**: Thats a weird-*** phone number regex.

Examples of propper phone numbers acording to this regex:
`+++++++++++++++1`
`              1`
`1+1          1+1`
`69`
`0`
`420`

If we don't enforce phone numbers at all, then at least make a propper phone regex.

implementation/jobberknoll/app/use-cases/register-use-case.test.ts
```diff
@@ -0,0 +1,80 @@
+import { accountMock, isErr, isOk, isSome, uuid } from "@jobberknoll/core/shared";
```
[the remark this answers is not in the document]
mlodybercik: drool

implementation/jobberknoll/app/use-cases/register-use-case.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export class RegisterUseCase extends UseCase<RegisterReq, Account, InvalidAccountDataError> {
+  public constructor(logger: Logger, private readonly accountRepo: AccountRepo) {
+    super(logger);
+  }
+
+  protected async handle(ctx: Ctx, req: RegisterReq): Promise<Result<Account, InvalidAccountDataError>> {
+    if (await this.accountRepo.isEmailTaken(ctx, req.email)) {
+      return err(invalidAccountData("email"));
+    }
```
mlodybercik, line 22: thought: Shouldn't be there other error type just for email? User tries to register (let's suppose he/she/them, forgot), and receives message that there is something wrong with email; what is? Same could apply for other fields, or at least for password strength. Frontend would validate before sending, but API is API.

implementation/jobberknoll/infra/account-repo/postgres-account-repo.ts
```diff
@@ -84,6 +84,19 @@ export class PostgresAccountRepo extends AccountRepo {
...
+    await client.queryObject`
+      UPDATE account SET
+        type = ${account.type},
+        full_name = ${account.fullName},
+        email = ${account.email},
+        hashed_password = ${account.hashedPassword},
+        last_modified = ${account.lastModified},
+        phone_number = ${"phoneNumber" in account ? account.phoneNumber : null}
+      WHERE id = ${account.id}`;
+  }
```
mlodybercik, line 98: Nice ORM, did you make it yourself? (lol)

implementation/jobberknoll/tests/delete-self-e2e.test.ts
```diff
@@ -0,0 +1,62 @@
+import { newCtx, UserRole } from "@jobberknoll/app";
```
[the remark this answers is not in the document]
mlodybercik: tF_Fkl

implementation/jobberknoll/tests/edit-account-password-e2e.test.ts
```diff
@@ -0,0 +1,104 @@
+import { newCtx, UserRole } from "@jobberknoll/app";
```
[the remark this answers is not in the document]
mlodybercik: dr-house-house

mlodybercik (commented):
Does edit-account need multiple endpoints? I think all of them could be done as single one that changes acording to user permissions.

implementation/jobberknoll/infra/account-repo/postgres-account-repo.ts
tchojnacki: Yes.

implementation/jobberknoll/app/shared/schemas.ts
tchojnacki: Will fix alongside password according to our messenger talk.

tchojnacki:
> Does edit-account need multiple endpoints? I think all of them could be done as single one that changes acording to user permissions.

I believe that it would highly break the single responsibility principle in this case, since these cases are pretty different:
- permissions: `/name` and `/password` are for `member`s, `phone` is for `passenger`s
- external dependencies: `/password` uses email and hashing, `/name` doesn't use anything, `/phone` would ideally send an SMS to confirm
- audit logging: each of those produces a different audit log (which is crucial for security in the case of `/password`, I would have to `if`/`switch` through what's changed)
- validation: `/password` requires additional validation outside of the DTO validator, which would have to be optionally `if`-ed out
- authentication: `/password` has to invalidate refresh tokens, while the other ones don't
- optionality: `phoneNumber` can be set as well as cleared, if it is missing in the request should we treat it as clearing or not-modifying?

We would end up with a monster endpoint, which:
- may or may not log you out
- may produce 0, 1, 2 or 3 audit logs
- may or may not send a mail
- may or may not send an SMS
- may require you to have a certain role depending on the request body

And a monster implementation which would basically have the current three use case codes combined into one and `if`ed out.

implementation/jobberknoll/app/use-cases/register-use-case.ts
tchojnacki: > that there is something wrong with email; what is

If the email is invalid, they would get a `schema-mismatch`, `invalid-account-data` is only returned if the email is taken (although I agree that the name could be better), since this is the only check which the frontend can't make.

> Frontend would validate before sending, but API is API

I believe that since it isn't an API for consumers and only for use in our system there's no need to return more details from the API, since they would only be needed in some way if the frontend is implemented incorrectly (and in such case, logs can be used to debug).

It would be a waste of time to do detailed password reporting such as "the password lacks an uppercase letter" instead of "password does not match the pattern", since the frontend won't use this info anyways (password requirements should be displayed to the user before any request to the backend gets made).

mlodybercik (approved):
Well done!

PR #63, opened 2025-01-13
PR: Setup Api Gateway

piterek130:


implementation/phoenix/src/main/java/com/example/phoenix/AddRequestIdFilter.java
jakubzehner, line 1: Since you are using spring it would be better to use an interceptor than a filter, since the filter is not part of it, but the web server that spring uses. For implementation details you can look at clabbert.

image

implementation/phoenix/src/main/java/com/example/phoenix/AddRequestIdFilter.java
```diff
@@ -0,0 +1,30 @@
...
+public class AddRequestIdFilter extends AbstractGatewayFilterFactory<AddRequestIdFilter.Config> {
+
+    public AddRequestIdFilter() {
+        super(Config.class);
+    }
+
+    @Override
+    public GatewayFilter apply(Config config) {
+        return (exchange, chain) -> {
+            String requestId = UUID.randomUUID().toString();
```
jakubzehner: This is not the UUIDv7 we agreed to use.

jakubzehner (changes requested):

implementation/phoenix/src/main/java/com/example/phoenix/AddRequestIdFilter.java
tchojnacki: As discussed, this should be ignored ⬆️

implementation/phoenix/.gitattributes
```diff
@@ -0,0 +1,2 @@
+/mvnw text eol=lf
+*.cmd text eol=crlf
```
tchojnacki, line 2: **nit:** This is mostly unrelated to the PR at hand, but it bugged me in the last and current PRs.

In `.gitattributes` you can mark files as vendored (meaning "not written by you") like this: https://github.com/github-linguist/linguist/blob/main/docs/overrides.md#vendored-code. It then doesn't count such files towards repository statistics and automatically collapses them in "Files changed" of the PR.

This treatment would be perfect for `mvnw`, `mvnw.cmd` and `maven-wrapper.properties`, which were not written by you and will never be edited.

implementation/phoenix/src/main/resources/application.yml
```diff
@@ -0,0 +1,48 @@
...
+    gateway:
+      routes:
+        - id: accounts_route
+          uri: ${JOBBERKNOLL_IP}
+          predicates:
+            - Path=/phoenix/account/**
+          filters:
+            - RewritePath=/phoenix/account/(?<remaining>.*), /ext/${remaining}
+            - AddRequestIdFilter
+            - AddRequestHeader=user-agent, Phoenix/1.0.0
```
tchojnacki: **nit:** Would be cool to take the `1.0.0` from a variable or something, to not have to change it in 4 separate places during a version bump.

implementation/phoenix/src/main/resources/application.yml
```diff
@@ -0,0 +1,48 @@
...
+    org.springframework.cloud.gateway: DEBUG
+
+spring:
+  config:
+    import: classpath:env.properties
+  cloud:
+    gateway:
+      routes:
+        - id: accounts_route
+          uri: ${JOBBERKNOLL_IP}
```
tchojnacki: **issue:** https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables, as discussed during Discord call.

implementation/phoenix/src/main/resources/application.yml
```diff
@@ -0,0 +1,48 @@
...
+          predicates:
+            - Path=/phoenix/account/**
+          filters:
+            - RewritePath=/phoenix/account/(?<remaining>.*), /ext/${remaining}
```
tchojnacki:
```suggestion
            - Path=/api/account/**
          filters:
            - RewritePath=/api/account/(?<remaining>.*), /ext/${remaining}
```

**issue:** As discussed during Discord call.

jakubzehner (approved):

implementation/phoenix/.gitattributes
```diff
@@ -0,0 +1,6 @@
...
+
+mvnw -linguist-vendored
+mvnw.cmd -linguist-vendored
+maven-wrapper.properties -linguist-vendored
```
tchojnacki: **issue:** I think you just **UN**-vendored them? Because of the leading `-`. I interpret the linked example as:
```
special-vendored-path/* linguist-vendored  # Mark as vendored
jquery.js -linguist-vendored               # Remove the mark
ano-dir/** linguist-vendored               # Mark as vendored
```

```suggestion
mvnw linguist-vendored
mvnw.cmd linguist-vendored
maven-wrapper.properties linguist-vendored
```

That's probably why the files are still visible in my view.

implementation/phoenix/src/main/java/com/example/phoenix/UUIDv7Generator.java
```diff
@@ -0,0 +1,14 @@
...
+
+import com.fasterxml.uuid.Generators;
+import com.fasterxml.uuid.impl.TimeBasedEpochRandomGenerator;
+
+public class UUIDv7Generator {
+
+    private static final TimeBasedEpochRandomGenerator UUID_GENERATOR =
+            Generators.timeBasedEpochRandomGenerator();
+
+    public String generate() {
```
tchojnacki: **suggestion:** This can also be static, since it doesn't use any instance members.

And if it is made static, we can remove `private final UUIDv7Generator uuidv7Generator;` and `this.uuidv7Generator = new UUIDv7Generator();` and use `String requestId = Uuidv7Generator.generate();` directly.

tchojnacki (approved):

implementation/phoenix/src/main/java/com/example/phoenix/AddRequestIdFilter.java
```diff
@@ -0,0 +1,28 @@
+package com.example.phoenix;
+
+import org.springframework.cloud.gateway.filter.GatewayFilter;
+import org.springframework.cloud.gateway.filter.factory.AbstractGatewayFilterFactory;
+import org.springframework.stereotype.Component;
+import org.springframework.web.server.ServerWebExchange;
+
+@Component
+public class AddRequestIdFilter extends AbstractGatewayFilterFactory<AddRequestIdFilter.Config> {
```
jakubzehner, line 9: I'm still not sure if this is a good solution, the interceptor here was obviously wrong as we agreed, but `AbstractGatewayFilterFactory<AddRequestIdFilter.Config>` doesn't convince me very much. I found [something like this](https://docs.spring.io/spring-cloud-gateway/reference/spring-cloud-gateway/global-filters.html), looks more reasonable. Anyway, you don't need to change it if it works well.

implementation/phoenix/src/main/resources/application.yml
```diff
@@ -0,0 +1,48 @@
...
+            - AddRequestHeader=user-agent, Phoenix/${PHOENIX_VERSION}
+
+        - id: logistics_route
+          uri: http://${LEPRECHAUN_SERVICE_HOST}:${LEPRECHAUN_SERVICE_PORT}
+          predicates:
+            - Path=/api/logistics/**
+          filters:
+            - RewritePath=/api/logistics/(?<remaining>.*), /ext/${remaining}
+            - AddRequestIdFilter
+            - AddRequestHeader=user-agent, Phoenix/${PHOENIX_VERSION}
```
jakubzehner, line 48: I don't remember our agreements, but it makes the paths look like this, for example:
- `/api/ticket/v1/tickets`
- `/api/ticket/v1/offers`
- `/api/account/v1/accounts`

Is that what it was supposed to look like?
@piterek130 @tchojnacki @mlodybercik

jakubzehner (approved):

implementation/phoenix/src/main/java/com/example/phoenix/AddRequestIdFilter.java
piterek130: I mocked servers and after test i think its working well as below
image

implementation/phoenix/src/main/resources/application.yml
jakubzehner: It is, as discussed

mlodybercik (approved):
472823333_457439937421552_5250876966024084081_n

PR #65, opened 2025-01-14
PR: Password security

tchojnacki:


implementation/jobberknoll/Dockerfile
```diff
@@ -1,24 +1,13 @@
-FROM ubuntu:24.10 AS runner-base
```
tchojnacki, line 1: **note:** We have to use the Deno runtime, since standalone executable can't execute FFI to hash the passwords :C

implementation/jobberknoll/infra/account-repo/postgres-account-repo.ts
```diff
@@ -40,7 +40,7 @@ export class PostgresAccountRepo extends AccountRepo {
         type ACCOUNT_TYPE NOT NULL,
         full_name VARCHAR(255) NOT NULL,
         email VARCHAR(255) NOT NULL UNIQUE,
-        hashed_password CHAR(60) NOT NULL,
+        hashed_password VARCHAR(255) NOT NULL,
```
tchojnacki, line 43: **note:** While the length of all hashes computed by a key derivation function configured with certain options remains constant, it changes depending on the algorithm and its options. So the column should sadly support variable-length data, to make sure we don't have to do a database migration when e.g. increasing the work factor.

documentation/adrs/003-password-hashing.md
```diff
@@ -0,0 +1,33 @@
...
+- Using a less secure key derivation function (e.g. bcrypt)
+
+## Decision Outcome
+
+Chosen option: **Using Argon2id with FFI**, because it heavily simplifies the service implementation while retaining almost indistinguishable functionality for the consumers.
+
+The backend must use Argon2id implemented in a native language (e.g. Rust, C, C++) and accessed via FFI. After auditing circa 10 libraries, it was decided to use [@node-rs/argon2](https://www.npmjs.com/package/@node-rs/argon2), which provides bindings for RustCrypto. It is the most popular library for Argon2id in the Node.js package registry. It is well supported in the Deno runtime and provides the `verify` function, which offloads the password verification from our service (and the process is not trivial, since it requires a constant-time comparison). The OWASP recommendations for algorithm parameters should be followed.
+
+The password must be stored using the [modular PHC string format](https://github.com/P-H-C/phc-string-format/blob/master/phc-sf-spec.md), to retain the ability to change the hashing algorithm in the future. Passwords should be rehashed on first login after the change of the hashing algorithm.
```
mlodybercik: **question**: Will you implement it tho?

implementation/jobberknoll/Dockerfile
mlodybercik, line 1: Nuh-uh... You can still do it with `deno compile`!!!

As discussed in PM's you can use FFI and call `argon2.so` with propper variables! Only requirement is to have `libargon2` installed because deno can not compile it into its binary (see [here](https://github.com/denoland/deno/issues/23266)). This means we need a distro that has them precompiled (preferably those listed [here](https://pkgs.org/search/?q=libargon2))
PoC:
```TypeScript
const argon2i_hash_encoded_params = {
  parameters: [
    /* const uint32_t t_cost */ <Deno.NativeNumberType>"u32",
    /* const uint32_t m_cost */ <Deno.NativeNumberType>"u32",
    /* const uint32_t parallelism */ <Deno.NativeNumberType>"u32",
    /* const void *pwd */ <Deno.NativeBufferType>"buffer",
    /* const size_t pwdlen */ <Deno.NativeBigIntType>"usize",
    /* const void *salt */ <Deno.NativeBufferType>"buffer",
    /* const size_t saltlen */ <Deno.NativeBigIntType>"usize",
    /* const size_t hashlen */ <Deno.NativeBigIntType>"usize",
    /* char *encoded */ <Deno.NativeBufferType>"buffer",
    /* const size_t encodedlen */ <Deno.NativeBigIntType>"usize",
  ],
  result: <Deno.NativeNumberType>"i32",
};

const dlib = Deno.dlopen("libargon2.so.1", {
  argon2i_hash_encoded: argon2i_hash_encoded_params,
});

const HASH_LEN = 32;
const TIME_COST = 2;
const MEMORY_COST = 1024;
const PARALLELISM = 1;

function hashPasswordEncoded(password: string, salt: string) {
  const pwd = new TextEncoder().encode(password);
  const saltBuf = new TextEncoder().encode(salt);
  const encoded = new Uint8Array(256);

  dlib.symbols.argon2i_hash_encoded(
    TIME_COST,
    MEMORY_COST,
    PARALLELISM,
    pwd,
    pwd.length,
    saltBuf,
    saltBuf.length,
    HASH_LEN,
    encoded,
    256
  );

  return new TextDecoder().decode(encoded);
}

console.log(hashPasswordEncoded("password", "very-long-salt"));
// same as: `echo -n 'password' | argon2 'very-long-salt' -i -t 2 -k 1024 -p 1 -l 32 -e`

dlib.close();
```

mlodybercik (changes requested):

implementation/jobberknoll/Dockerfile
mlodybercik: Better yet! We can get it to work even on Alpine Linux if we want! 
obraz

implementation/jobberknoll/Dockerfile
```diff
@@ -1,13 +1,24 @@
-FROM denoland/deno:2.1.4
-
-WORKDIR /app
+FROM ubuntu:24.10 AS runner-base
 
 RUN apt-get -y update
-RUN apt-get -y install curl
+RUN apt-get -y install curl libargon2-1
```
mlodybercik, line 4: As discussed in PM's, this leaves litter in images. I'll optimize this out soon™

implementation/jobberknoll/app/security/libargon2.ts
```diff
@@ -0,0 +1,75 @@
+// SAFETY: see ADR-003
+
+// SAFETY: even if the FFI fails, the code will error out, so the user will get a 5xx error, which is bad, but safe
+const libargon2 = Deno.dlopen(
```
mlodybercik, line 4: Technically, when stopping, one should call `libargon2.close()` (see [here](https://man7.org/linux/man-pages/man3/dlopen.3.html)) but I don't think its neccessary in docker images. We should keep it in mind tho.

documentation/adrs/003-password-hashing.md
tchojnacki: Updated.

implementation/jobberknoll/Dockerfile
tchojnacki: Gooooood job 🎆

mlodybercik (approved):
🎊
high-five-cat

PR #74, opened 2025-01-17
PR: Set up JWT handling

tchojnacki:
obraz
obraz
obraz


mlodybercik (approved):
My only nits are:
- do you really have to generate new ECDSA key pair for each test (it takes next to no time, but still, tests should be rapid)
- keys *should not* be placed in .env as plain text, rather they should be read from some secret storage. As an court-appointed DevOps I'll let that slide for now, as I did not specify how secrets should be managed.


e00a5d51ffa516ef2b64ea1ff3a3a60a

tchojnacki:
> * do you really have to generate new ECDSA key pair for each test (it takes next to no time, but still, tests should be rapid)

I had a single keypair hardcoded for ~90% of this PRs lifetime, but then decided that if I have some keys laying around in the repository then some not-so-smart person (me) might try to use them in staging or prod for convenience. So I now generate them to 1. make sure the tests are keypair-agnostic 2. make sure these keys won't get reused.

> * keys _should not_ be placed in .env as plain text, rather they should be read from some secret storage. As an court-appointed DevOps I'll let that slide for now, as I did not specify how secrets should be managed.

Agreed. This also applies to DB connection etc.

PR #75, opened 2025-01-18
PR: `leprechaun` internal endpoints, filters, treating DTO's as interfaces

mlodybercik:


implementation/leprechaun/src/vehicle/test/vehicle-filter.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+      new Vehicle('1234'),
+      new Vehicle('2345'),
+      new Vehicle('3456'),
+      new Vehicle('4567'),
+      new Vehicle('5678'),
+    ];
+
+    await em.persistAndFlush(vehicles);
+
+    testCase = async (filter: VehicleFilterOptions, expected: Array<number>) => {
```
tchojnacki:
```suggestion
    testCase = async (filter: VehicleFilterOptions, expected: number[]) => {
```

implementation/leprechaun/src/vehicle/controller/vehicle.dto.ts
```diff
@@ -14,18 +14,21 @@ export class VehicleDto extends GenericIdActiveDto {
...
 
   static fromEntity(entity: Vehicle): VehicleDto {
-    return new VehicleDto(entity.id, entity.sideNumber, entity.isActive);
+    const dto = new VehicleDto();
+    Object.assign(dto, <VehicleDto>{
+      id: entity.id,
+      sideNumber: entity.sideNumber,
+      isActive: entity.isActive,
+    });
+    return dto;
```
tchojnacki:
```suggestion
    return Object.assign(new VehicleDto(), <VehicleDto>{
      id: entity.id,
      sideNumber: entity.sideNumber,
      isActive: entity.isActive,
    });
```

implementation/leprechaun/src/shared/api/filter.ts
```diff
@@ -0,0 +1,28 @@
...
+    forbidUnknownValues: true, // remove unknown values
+    forbidNonWhitelisted: true, // remove non-whitelisted properties
+    stopAtFirstError: false, // continue validation after first error
+  }).map((error) => delete copy[error.property as keyof T]);
+  return copy;
+}
+
+// https://www.30secondsofcode.org/js/s/escape-reg-exp/
+export const escapeRegExp = (str: string) => str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+export const EscapedRegExpLike = (str: string) => RegExp(escapeRegExp(str));
```
tchojnacki, line 28: Sounds like a security AND performance nightmare. Can't we do something [like this](https://mikro-orm.io/docs/query-conditions#postgresql)?

implementation/leprechaun/src/shared/test/helpers.ts
```diff
@@ -1,22 +1,66 @@
...
+};
+
+export type FuncOrPromise<T = unknown> = (() => Promise<T>) | Promise<T>;
+
+export function createTimeOffsetFromNow({ hours, seconds }: TimeType): Date {
+  if (seconds === undefined) {
+    seconds = 0;
+  }
   return new Date(new Date().valueOf() + (3600 * hours + seconds) * 1000);
 }
```
tchojnacki:
```suggestion
export function createTimeOffsetFromNow({ hours, seconds = 0 }: TimeType): Date {
  return new Date(new Date().valueOf() + (3600 * hours + seconds) * 1000);
}
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#default_value

implementation/leprechaun/src/shared/api/filter.ts
```diff
@@ -0,0 +1,28 @@
+import { Type } from '@nestjs/common';
+import { ApiQueryOptions } from '@nestjs/swagger';
+import { plainToInstance } from 'class-transformer';
+import { validateSync } from 'class-validator';
+
+// TODO: there is a way to force TypeScript to check if the array of objects matches EXACTLY the interface
+// source? came up in a dream
+export type ForceArrayToHaveNamedApiQueryOf<T extends {}> = Array<{ name: keyof T } & ApiQueryOptions>;
```
tchojnacki, line 8:
```TS
type Exact<TExpected, TActual extends TExpected> = TExpected extends TActual ? TExpected: never;

type Foo = { x: number };
function acceptT<T extends Foo>(input: Exact<Foo, T>) {}

const x1 = {x: 1};
acceptT(x1); // OK

const x2 = {x: 1, y: 1};
acceptT(x2); // bad

const x3 = {y: 1};
acceptT(x3);  // bad

const x4 = {x: null};
acceptT(x4); // bad

const x5 = {};
acceptT(x5); // bad
```

But I don't know why you'd want this, it doesn't change anything in terms of runtime. If you actually want to remove the extra fields during runtime, something like [z.strict()](https://zod.dev/?id=strict) from Zod works fine.

implementation/leprechaun/src/route/test/route.controller.spec.ts
```diff
@@ -254,14 +304,14 @@ describe('RouteService', () => {
...
-    const c2e = createTimeOffsetFromNow(7);
-    const c3e = createTimeOffsetFromNow(8);
+    const c1s = createTimeOffsetFromNow({ hours: 1, seconds: 0 });
+    const c2s = createTimeOffsetFromNow({ hours: 2, seconds: 0 });
+    const start = createTimeOffsetFromNow({ hours: 3, seconds: 0 });
+    const c1e = createTimeOffsetFromNow({ hours: 4, seconds: 0 });
+    const c3s = createTimeOffsetFromNow({ hours: 5, seconds: 0 });
+    const end = createTimeOffsetFromNow({ hours: 6, seconds: 0 });
+    const c2e = createTimeOffsetFromNow({ hours: 7, seconds: 0 });
+    const c3e = createTimeOffsetFromNow({ hours: 8, seconds: 0 });
```
tchojnacki: If seconds are optional and equal to 0, why pass them?

tchojnacki (approved):
Scary

implementation/leprechaun/src/vehicle/controller/vehicle.dto.ts
mlodybercik: Did the same for all DTO's.

implementation/leprechaun/src/shared/api/filter.ts
mlodybercik: The issue with fulltext search here is that none of the values of `line.names`, `stop.names`, `sideNumbers` would make any sense to fulltext indexer as it depends on some structure of indexed text. Changing table to be used with indexes (as per your link) would brake `VehicleService`, and breaks some DTOs. Also, fulltext search is weird, it does not return values you'd hope for.

implementation/leprechaun/src/shared/test/helpers.ts
mlodybercik: Destructuring statements are still magic to me.

implementation/leprechaun/src/route/test/route.controller.spec.ts
mlodybercik: I did regex replace all `createTimeOffsetFromNow` and wrote it like that.

mlodybercik: I'll just `createTimeOffsetFromNow\(\{ hours: (\d+), seconds: 0 \}\)` into `createTimeOffsetFromNow({ hours: $1 })`

implementation/leprechaun/src/shared/api/filter.ts
mlodybercik: Nah, it is not for the runtime, I just want to make sure that all fields are being described in ApiQuery in filters.

mlodybercik: But the example you provided is not the thing I am looking for. It will be better if I describe the problem with an example:

```TypeScript
type ForceArrayFields<T> = (... & ApiQueryOptions)[];
interface SomeFields {
  a: string,
  b: string,
  c: string,
}

const array1: ForceArrayFields<SomeFields> = [
  { name: 'a', /* ... rest of the ApiQueryOptions */},
  { name: 'b', /* ... rest of the ApiQueryOptions */},
  { name: 'c', /* ... rest of the ApiQueryOptions */},
] // OK

const array2: ForceArrayFields<SomeFields> = [
  { name: 'a', /* ... rest of the ApiQueryOptions */},
] // NOT OK, missing 'b' and 'c'

const array3: ForceArrayFields<SomeFields> = [
  { name: 'a', /* ... rest of the ApiQueryOptions */},
  { name: 'b', /* ... rest of the ApiQueryOptions */},
  { name: 'c', /* ... rest of the ApiQueryOptions */},
  { name: 'd', /* ... rest of the ApiQueryOptions */},
] // NOT OK, has 'd' which is not in SomeFields

const array4: ForceArrayFields<SomeFields> = [
  { name: 'a', /* ... rest of the ApiQueryOptions */},
  { name: 'a', /* ... rest of the ApiQueryOptions */},
  { name: 'c', /* ... rest of the ApiQueryOptions */},
] // NOT OK, has 'a' twice

```

PR #78, opened 2025-01-22
PR: Inferius Endpoints

piterek130:


implementation/inferius/src/main/java/pwr/jakprzyjade/inferius/creditcardinfo/database/CreateCreditCardDto.java
```diff
@@ -0,0 +1,28 @@
...
+@Data
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+public class CreateCreditCardDto {
+    @Size(max = 255)
+    private String label;
+
+    @NotNull
+    @Size(min = 16, max = 16)
```
tchojnacki, line 21: **nit:** Can we ensure it only contains digits?

implementation/inferius/src/main/java/pwr/jakprzyjade/inferius/creditcardinfo/service/CreditCardInfoService.java
```diff
@@ -1,4 +1,94 @@
...
+        creditCardInfoRepository.delete(card);
+    }
+
+    public CreditCardDto addCreditCard(UUID userId, CreateCreditCardDto createDto) {
+        Wallet wallet = walletRepository.findByPassengerId(userId)
+                .orElseThrow(() -> new ResourceNotFoundException("Wallet not found for user ID: " + userId));
+
+        CreditCardInfo newCard = CreditCardInfo.builder()
+                .label(createDto.getLabel())
+                .number(createDto.getNumber())
```
tchojnacki, line 78: **issue:** Ideally, you'd want to validate it with Luhn's algorithm
obraz

But we can pretend, that you don't have to.

implementation/inferius/src/main/java/pwr/jakprzyjade/inferius/creditcardinfo/service/CreditCardInfoService.java
```diff
@@ -1,4 +1,94 @@
...
+                .holderName(createDto.getHolderName())
+                .expirationDate(createDto.getExpirationDate())
```
tchojnacki, line 80: **issue:** Ideally, the card should not be expired, but it's not even in the requirements, so we can skip that.

implementation/inferius/src/main/java/pwr/jakprzyjade/inferius/fine/database/FineCreateDto.java
```diff
@@ -0,0 +1,35 @@
...
+    @JsonDeserialize(using = UUIDv7Deserializer.class)
+    private UUID passengerId;
+
+    @NotNull
+    @Size(min = 1, max = 255)
+    private String recipient;
+
+    @NotNull
+    @Min(0)
+    private BigDecimal amountPln;
```
tchojnacki: **issue:** These should be probably in Grosze and only then mapped to PLN. JSON doesn't support decimal data types and money shouldn't be stored in floating point datatypes, so storing it as an integer in the smallest unit possible is really the only solution.

You can look at @jakubzehner's code, it is somewhere there (ticket price?).

tchojnacki (approved):

implementation/inferius/src/main/java/pwr/jakprzyjade/inferius/fine/database/FineCreateDto.java
jakubzehner: [Dto](https://github.com/PWR-ACS-SE-24/SoftwareSystemDesign/blob/d14889ba3b2cc8d4331664eecd4db5ffc024e50a/implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/contracts/ticketoffer/TicketOfferDto.java#L48)
[Mapper](https://github.com/PWR-ACS-SE-24/SoftwareSystemDesign/blob/d14889ba3b2cc8d4331664eecd4db5ffc024e50a/implementation/clabbert/src/main/java/pwr/jakprzyjade/clabbert/api/mappers/TicketOfferMapper.java#L74)

PR #79, opened 2025-01-23
PR: `DevOps` basic infrastructure implementation

mlodybercik:
Default student account sucks. Acording to E2, project should be hosted on a EKS ~~but account doesn't have required roles.~~
Issues encountered in no particular order:
- Nodes on EC2 could not join EKS cluster,
- nodes on EC2 could not join EKS cluster with SG set to allow all ingress.
- nodes on fargate could not join EKS cluster due to ¯\\\_(ツ)_/¯,
- student account does not have required roles to administer a cluster by default,
- student account can use at most `large` machines.
- minikube has `--driver=ssh` option that does not work (creates docker config that colides with itself)
- oh yeah, did I mention that voclabs account sucks?
- ~~I am slowly lossing my mind.~~

I will make it work _eventually_... Acording to my friend (a real one, I promise), settings done in generic kubernetes cluster are mostly portable, ~~and if it weren't for the issues with AWS they would work the same as they do on `minikube`.~~

EDIT 1:  
Nevermind! I got it to work! But I will not delete the previous paragraph, because it is a testament to my struggle and hours of my life that I will never get back.
The issue with previous clusters were that I was using autoEKS, which is not exactly supported by our student roles? Nonetheless, I created a new cluster with some help from a colleague and it works! (thank you Paweł, you are a lifesaver 💝)

Some notes:
- We won't do full E2 plan, because our account is too limited to do so.
- Why did I choose public subnets instead of NAT + multiple private subnets? 
  Costwise (for smaller clusters), it is cheaper to use public subnets. 
  
  Let's assume three AZ's, each with one public subnet, and three private networks with three NAT gateways. Its cheaper to use public subnets if amount of nodes in a cluster is less than 18 (roughly 9 nodes per 1 NAT gateway, not counting data transfer).


tchojnacki (approved):

implementation/leprechaun/src/config/config.ts
```diff
@@ -9,7 +9,7 @@ const config: AppConfig = {
   PROD: process.env.NODE_ENV === 'production',
   LEPRECHAUN_APP_PORT: parseInt(process.env.LEPRECHAUN_APP_PORT!) || 3000,
   LEPRECHAUN_DATABASE_HOST: process.env.LEPRECHAUN_DATABASE_HOST ?? 'postgresql://postgres:test@localhost:5432',
-  LEPRECHAUN_DATABASE_DBNAME: process.env.LEPRECHAUN_DATABASE_HOST ?? 'leprechaun',
+  LEPRECHAUN_DATABASE_DBNAME: process.env.LEPRECHAUN_DATABASE_DBNAME ?? 'leprechaun',
```
mlodybercik, line 12: Oops :p

tchojnacki (approved):

PR #80, opened 2025-01-23
PR: Adjust libargon dylib opening

tchojnacki:
This solves two issues:
- the dynamic library is now properly closed after usage (@mlodybercik reported this issue before, but we decided to ignore it for now)
- `libargon2` is not required to use Jobberknoll as a library (as opposed to binary), so namely, Feather no longer needs `libargon2` or `--allow-ffi`


jakubzehner (approved):
i dont understand but looks fine

PR #83, opened 2025-01-23
PR: `clabbert` change `Dockerfile` to improve build times

### LINKED ISSUE #82: `clabbert` long build times
Clabbert takes long time to build, around 3 minutes. It could be <30s.

https://github.com/PWR-ACS-SE-24/SoftwareSystemDesign/blob/main/implementation/clabbert/Dockerfile#L6-L7

mlodybercik:
180s to 13.5s 🎉

Closes #82


jakubzehner (approved):
In my case, it does not reduce the time a bit, but it takes up more space. Great job. (Maybe possibly it reduces the time of the first build, but I don't feel like checking so precisely).

tchojnacki (approved):

PR #84, opened 2025-01-23
PR: Phoenix authorization filter

piterek130:


piterek130:
image

jakubzehner (commented):
It looks like it should work, but the errors you're generating don't follow the format we've agreed upon, it's hard to tell at all what they look like just by looking at the code.

In addition, what happens in a situation where someone already attaches our `jp` headers to a request?

And also looking at your screenshot should the response include `jp-user-id` and `jp-user-role` headers?

If the response goes through Gateway going back to the client shouldn't it contain `user-agent` `phoenix`?

I will wait with approval for someone else's opinion.
@tchojnacki @mlodybercik

implementation/inferius/README.md
```diff
@@ -0,0 +1,49 @@
...
+| `PUT    /ext/v1/fines/:id/cancel`  | 🟩         |
+| `GET    /int/v1/health`            | 🟩         |
+| `GET    /int/v1/health/:component` | 🟩         |
+| `GET    /int/v1/endpoints`         | 🟥         |
+
+## Infrastructure
+
+| **Integration**             | **Status** |
+| --------------------------- | ---------- |
+| Database (PostgreSQL)       | 🟩         |
```
tchojnacki: **nit:** Shouldn't this include:
- integration with Kuba's service
- external payment gateway (t-pay)
(and possibly my service, but it can be skipped as discussed)

implementation/phoenix/compose.yml
```diff
@@ -16,4 +16,6 @@ services:
       - INFERIUS_SERVICE_HOST=host.docker.internal
       - INFERIUS_SERVICE_PORT=1082
       - LEPRECHAUN_SERVICE_HOST=host.docker.internal
-      - LEPRECHAUN_SERVICE_PORT=1083
\ No newline at end of file
+      - LEPRECHAUN_SERVICE_PORT=1083
+      - FEATHER_SERVICE_HOST=localhost
+      - FEATHER_SERVICE_PORT=8001
```
tchojnacki, line 21: **note:** Since Feather would be Phoenix's sidecar, [and thus run as a second container in the same pod](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/), there's no need to follow the Kubernetes convention here (but you still can).

implementation/phoenix/src/main/java/com/example/phoenix/AuthorizationFilter.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Override
+    public GatewayFilter apply(Config config) {
+        return (exchange, chain) -> {
+            String authorizationHeader = exchange.getRequest().getHeaders().getFirst(HttpHeaders.AUTHORIZATION);
+
+            if (authorizationHeader == null || !authorizationHeader.startsWith("Bearer ")) {
+                return Mono.error(new RuntimeException("Missing or invalid Authorization header"));
+            }
```
tchojnacki: **issue:** While invalid auth header should infact throw, a missing auth header is valid (and means that the user is a guest).

implementation/phoenix/src/main/java/com/example/phoenix/AuthorizationFilter.java
```diff
@@ -0,0 +1,90 @@
...
+
+            String token = authorizationHeader.substring(7);
+
+            Map<String, String> requestBody = Map.of("accessToken", token);
+
+            System.out.println(requestBody);
```
tchojnacki: **nit:** Should probably get removed after debugging.

implementation/phoenix/src/main/java/com/example/phoenix/AuthorizationFilter.java
```diff
@@ -0,0 +1,90 @@
...
+                                }))
+                                .build();
+                        return chain.filter(modifiedExchange);
+                    })
+                    .onErrorResume(e -> Mono.error(new RuntimeException("Authorization failed", e)));
+        };
+    }
+
+    public static class Config {
+    }
```
tchojnacki, line 109: **nit:** Is this necessary?

implementation/phoenix/src/main/resources/application.properties
```diff
@@ -1,2 +1,3 @@
 spring.application.name=phoenix
 phoenix.version=1.0.0
+server.port=8081
```
tchojnacki: **question:** Is this Phoenix's port? If so I'd avoid it being the same as Feather's.

tchojnacki:
> the errors you're generating don't follow the format we've agreed upon

Agree, sounds like a blocking issue.

> In addition, what happens in a situation where someone already attaches our `jp` headers to a request?

It is mentioned in the contracts _"Phoenix MUST discard well-known headers from the incoming requests (from outside the system) for security reasons. This behavior SHOULD be validated in the integration tests."_

> And also looking at your screenshot should the response include `jp-user-id` and `jp-user-role` headers?

It probably shouldn't, but there's also no harm if it does. The sender already knows the contents of both of these headers - since he has the `accessToken`, he can read its payload even without any keys or secrets or anything.

> If the response goes through Gateway going back to the client shouldn't it contain `user-agent` `phoenix`?

Yes, as mentioned in the contracts _"Phoenix SHOULD set the user-agent header to Phoenix/1.0.0 (or other version) for any outgoing request."_

tchojnacki:
The general logic is solid and correct, but there are some details to improve 👍

implementation/inferius/README.md
jakubzehner: With clabbert - no
But with sqs that clabbert also use - yes and it will stay red to the end

implementation/phoenix/src/main/java/com/example/phoenix/AuthorizationFilter.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    public GatewayFilter apply(Config config) {
+        return (exchange, chain) -> {
+            String authorizationHeader = exchange.getRequest().getHeaders().getFirst(HttpHeaders.AUTHORIZATION);
+
+            if (authorizationHeader == null || !authorizationHeader.startsWith("Bearer ")) {
+                return Mono.error(new RuntimeException("Missing or invalid Authorization header"));
+            }
+
+            String token = authorizationHeader.substring(7);
```
mlodybercik: **nit**: that *7* does not mean anything. How about you define something like this:
```java
private final static int AUTHORIZATION_HEADER_LENGTH = String("Bearer ").length();
...
String token = authorizationHeader.substring(AuthorizationFilter.AUTHORIZATION_HEADER_LENGTH);
```

tchojnacki (approved):

PR #88, opened 2025-01-24
Clabbert - Fix no header parameter in external openapi swagger

jakubzehner:
- fix no header parameter in external openapi swagger


mlodybercik (approved):
image

